PR #8, opened 2020-09-21
move database setup out of form component into context provider

micheal-w-wells:


micheal-w-wells:
redundant with switch to pouch

PR #13, opened 2020-09-23
Force single geom

popkinj:
Forcing only one geometry at a time.

### DIFF
--- a/app/src/components/map/MapContainer.css
+++ b/app/src/components/map/MapContainer.css
@@ -1,36 +1,3 @@
-.container {
-  text-align: center;
-  position: absolute;
-  left: 0;
-  right: 0;
-  top: 50%;
-  transform: translateY(-50%);
-}
-
-.container strong {
-  font-size: 20px;
-  line-height: 26px;
-}
-
-.container p {
-  font-size: 16px;
-  line-height: 22px;
-  color: #8c8c8c;
-  margin: 0;
-}
-
-.container a {
-  text-decoration: none;
-}
-
-#map {
-  height: 100%;
-  width: 100%;
-}
-
-.MuiContainer-root {
-  padding: 0 !important;
-}
 
 .leaflet-draw.leaflet-control {
   transform: scale(1.5) translate(-5px,40px);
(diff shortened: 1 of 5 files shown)


app/src/components/map/MapContainer.css
```diff
@@ -0,0 +1,45 @@
...
+
+#map {
+  height: 100%;
+  width: 100%;
+}
+
+.MuiContainer-root {
+  padding: 0 !important;
+}
+
```
NickPhura: Can you strip out this css? I tested it out, and It doesn't appear to be needed

All of the general container type css is being controlled by a parent component, and then passed down.  The goal is to slowly build up a single css theme object (got the bare bones going currently), that can then be referenced everywhere, so the app pages all have the same look but without having to maintain css in multiple places.

NickPhura (dismissed):
Just had the one comment about the extra css.

Otherwise looks good!

app/src/components/map/MapContainer.css
rstens: Failing build: 
```
react-scripts] H:/invasivesbc/app/src/App.tsx
[react-scripts] TypeScript error in H:/invasivesbc/app/src/App.tsx(5,48):
[react-scripts] Cannot find module 'contexts/DatabaseChangesContext'.  TS2307
[react-scripts]     3 | import { CircularProgress, makeStyles } from '@material-ui/core';
[react-scripts]     4 | import { KeycloakProvider } from '@react-keycloak/web';
[react-scripts]   > 5 | import { DatabaseChangesContextProvider } from 'contexts/DatabaseChangesContext';
[react-scripts]       |                                                ^
[react-scripts]     6 | import Keycloak, { KeycloakConfig, KeycloakInstance } from 'keycloak-js';
[react-scripts]     7 | import React from 'react';
[react-scripts]     8 | import AppRouter from './AppRouter';
```

popkinj: Right... I asked about this earlier in Teams. This css is needed for the map. And I'll need to add further in the future. Is there a better way to add css to the map?

popkinj: Thanks for tracking that down @rstens . But I don't think this PR touches that file.

popkinj (commented):
Just removed the offending code that @rstens pointed out.

NickPhura (approved):

PR #45, opened 2020-09-30
Offline map

popkinj:
Geometries persist offline


rstens (commented):
So why did you go to an older version of leaflet-draw?

rstens (changes requested):
There are some complaints about no new-lines at the end of several files.

rstens (commented):
As a general comment, when I pull this branch down and built it locally, it was/is based on a rather old earlier version. This could cause issue when merging. So I'd like to suggest that you test the new code first with a rebased branch, so that we do not find issues later.

popkinj:
Good idea. I'll do a rebase.

popkinj:
Hey @NickPhura and @rstens  . Can one of you check this PR. I've just made a couple adjustments. Rebase and removing code smells.

rstens (approved):

popkinj:
Thanks @rstens :)

PR #72, opened 2020-10-08
Photos on web and device

anissa-agahchen:
* Got photos storing to DB and reading both on web and mobile devices
* removed dependencies on the filesystem and storage


rstens (changes requested):
Why remove these? Now the pipeline is failing as it expects these files.
You should not have to touch these.

These removal changes need to be eliminated from the PR because if we merge them if will affect the main repo.

rstens:
Please assign reviewers in this PR (top right in the side bar).
Also make sure that you rebase.
Lastly please review the new code smell by clicking in the SonarCloud code smell link

rstens (changes requested):
Please do not add the IOS and Android directories to the .gitignore.

You have to change this.

rstens:
Also there are some small updates to IOS and Android that are kept in the directories that will make them work (better), if these are eliminated we loose those small changes.

anissa-agahchen:
Need to fix changes that break the build

PR #75, opened 2020-10-08
Photo on device

anissa-agahchen:


app/src/components/photo/PhotoContainer.tsx
```diff
@@ -1,15 +1,86 @@
...
 const PhotoContainer: React.FC<IPhotoContainerProps> = (props) => {
-  const { photos, takePhoto, deletePhotos } = usePhotoGallery();
+  const { getPhoto } = useCamera();
+  const [photos, setPhotos] = useState<Photo[]>([]);
+  const databaseContext = useContext(DatabaseContext);
+
+  const initPhotos = async function (activityDoc: any) {
+    const activity_id = activityDoc._id;
+    const dbDoc = await databaseContext.database.get(activity_id)
+    const dbPhotos = dbDoc.photos != undefined ? dbDoc.photos : [];
```
NickPhura: (Mainly a subjective preference/something interesting if you havent seen this style before)

You could also say:

```javascript
const dbPhotos = dbDoc.photos || [];
```

If `dbDoc.photos` is not null or undefined, it will be assigned, otherwise it will be a falsy value and fallback to assigning the `[]`.

Useful if you ever need to do null checks at every level in a nested object:

```javascript
const someArray = parent && parent.child && parent.child.value || [];
```

The last item in the chain is returned if it is not falsy (in this case `parent.child.value`)

anissa-agahchen: nice, thank you

app/src/components/photo/PhotoContainer.tsx
```diff
@@ -1,15 +1,86 @@
...
+    const cameraPhoto = await getPhoto({
+      resultType: CameraResultType.Base64,
+      source: CameraSource.Camera,
+      quality: 100
+    });
+
+    const fileName = new Date().getTime() + '.jpeg';
+    const photo = {
+      filepath: fileName,
+      base64: "data:image/jpeg;base64," + cameraPhoto.base64String
```
NickPhura: Specifically in regards to `image/jpeg`, we probably don't want to hard code this, in case the file format is different (`image/png`, etc).  Can we infer the filetype from some property of the `cameraPhoto` object? 

Same for the file extension on line 51.

app/src/components/photo/PhotoContainer.tsx
```diff
@@ -1,15 +1,86 @@
...
+  };
+
+  const deletePhoto = async (filepath: any) => {
+
+    const reducedPhotos = photos.filter((photo) => photo.filepath != filepath);
+    setPhotos(reducedPhotos);
+  };
+
+  useEffect(() => {
+    if (props && databaseContext.database) {
```
NickPhura: Regarding `databaseContext.database`, we shouldn't need to check the database context for null anymore, as the app currently waits for the database to be setup before rendering anything.

NickPhura: Was easier to do it that way, rather than having to constantly account for a potentially null database in a ton of places.

NickPhura (commented):
Left a couple of comments, mostly just FYI's, nothing critical.

Looks good though!  Particularly like the use of a Photo interface :)

NickPhura (dismissed):

app/src/components/photo/PhotoContainer.tsx
anissa-agahchen: change made

NickPhura (approved):

PR #78, opened 2020-10-09
WIP trip cache page and map refactor

micheal-w-wells:
So far:

- get a paper, with a grid, and the map to render on the plan page
- upload kml component for plan trip page
- can db save a kml as geojson, and render points on map on load

Also some map refactoring to allow the component to do duty for more than one context

### DIFF
--- a/app/src/features/home/plan/PlanPage.tsx
+++ b/app/src/features/home/plan/PlanPage.tsx
@@ -6,7 +6,6 @@ import { DropzoneArea } from 'material-ui-dropzone';
 import React, { useContext, useEffect, useState } from 'react';
 
 import { kml } from '@tmcw/togeojson';
-var fs = require('fs');
 // node doesn't have xml parsing or a dom. use xmldom
 const DOMParser = require('xmldom').DOMParser;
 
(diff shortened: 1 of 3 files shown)


app/src/features/home/plan/PlanPage.tsx
```diff
@@ -1,15 +1,237 @@
...
+    height: '100%',
+    width: '100%'
+  },
+  heading: {
+    fontSize: theme.typography.pxToRem(18),
+    fontWeight: theme.typography.fontWeightRegular
+  },
+  layerPicker: {
+    height: '100%',
+    width: '100%'
```
NickPhura, line 64: Not an issue for now, but eventually we should be able to replace all of these `100%` with some flex-box css.  But we can organize the html/css later when things are stable

NickPhura (approved):

app/src/features/home/plan/PlanPage.tsx
micheal-w-wells: yep there's some quick and dirty CSS to deal with later for sure, thanks for taking a look!

Issue #46, opened 2020-10-01
offline basemap work


micheal-w-wells:
## Storing basemap for all zoom levels of user specified area
- [x] Select plugin
- [x] Select workflow (draw box... zoom to area etc...)
- [x] Connect plugin to storage mechanism (Webworker or PouchDB)
- [x] Widget / button to activate offline mechanism.
- [x] Integrate a timeout of stored basemap tiles

**Additional context**
Come up with plan for how to cache layers on mobile device. Subsequent ticket is #784
- [x] Plan for how to get layers onto the device (API)
- [x] Plan for storage medium on device (GRDB or ?)

PR #91, opened 2020-10-15
leaflet offline, jamies basemap caching and deleting, map refactor

micheal-w-wells:
* jamies tile caching, and map extent
* map refactor


micheal-w-wells:
#46

PR #101, opened 2020-10-19
Pipeline Tweaks

NickPhura:
- attempting to fix bad api route config
- Other bits of cleanup (removing unused values, etc)

### DIFF
--- a/api/.pipeline/config.js
+++ b/api/.pipeline/config.js
@@ -4,7 +4,7 @@ let options = require('pipeline-cli').Util.parseArguments();
 // The root config for common values
 const config = require('../../.config/config.json');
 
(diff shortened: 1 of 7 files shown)


micheal-w-wells:
i think this one broke the pipeline https://github.com/bcgov/invasivesbc/runs/1277623793

NickPhura:
> 
> 
> i think this one broke the pipeline https://github.com/bcgov/invasivesbc/runs/1277623793

The build was breaking before, these changes were trying to fix it, but evidently didn't work.
Something is up with the route it tries to create for the api.  It complains its in use already, so one of the names is colliding somewhere.

micheal-w-wells:
is this the race condition you were talking about?

PR #129, opened 2020-10-22
All SSO urls updated

rstens:


rstens:
Will merge even though the api test has a 401 failure.

PR #136, opened 2020-10-29
90: user/role table migration and seeding

NickPhura:
- Remove previously added migration that re-created all of the legacy lucy-web tables.
  - Because invasivesbc doesn't care about them (yet/maybe never?), and because they are fairly out of date now that we have a new codes set/value setup.
- Add migration to do initial db setup
  - In this case just creating the schema.
- Add migration for the application_user and user_role tables (same as lucy-web database)
- Add seed for test users and their roles (same as lucy-web)


rstens (commented):
Looks good

database/src/migrations/20200810113233_init.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function down(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set schema '${DB_SCHEMA}';
+    set search_path = ${DB_SCHEMA},public;
+
+    DROP SCHEMA IF EXISTS ${DB_SCHEMA}.invasivesbc;
```
anissa-agahchen: This is breaking the downgrade.  Just need to remove `.invasivesbc`
```suggestion
    DROP SCHEMA IF EXISTS ${DB_SCHEMA};
```

database/src/migrations/20200810113233_init.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set schema '${DB_SCHEMA}';
+    set search_path = ${DB_SCHEMA},public;
+
+    CREATE SCHEMA IF NOT EXISTS invasivesbc;
+
+    ALTER SCHEMA invasivesbc OWNER TO invasivebc;
```
anissa-agahchen:
```suggestion
    ALTER SCHEMA ${DB_SCHEMA} OWNER TO invasivebc;
```

database/src/migrations/20200810113233_init.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set schema '${DB_SCHEMA}';
+    set search_path = ${DB_SCHEMA},public;
+
+    CREATE SCHEMA IF NOT EXISTS invasivesbc;
```
anissa-agahchen:
```suggestion
    CREATE SCHEMA IF NOT EXISTS ${DB_SCHEMA};
```

anissa-agahchen (changes requested):
@NickPhura Looks really good.  
Just found one small breaking change on the migration downgrade, and put a couple of suggestions.

database/src/migrations/20200810113233_init.ts
NickPhura: Nice catch!

Also, never seen this 'commit suggestion' option before, thats interesting

anissa-agahchen (approved):

PR #140, opened 2020-10-30
load ref data properly

micheal-w-wells:


micheal-w-wells:
Sonar cloud bug is a false positive in this case, there's a commented out line that needs to be there until we fix an issue with the search endpoint not liking the geo feature.

PR #146, opened 2020-11-04
fix kml

micheal-w-wells:
- [x] id of uploaded kml geo lines up with trip record
- [ ] geo(s) from upload are visible without refreshing the page

For now, the first geo in the collection uploaded will be used in the search.  If
you upload one polygon, that will be the search boundary.


NickPhura:
Might just need to rebase this on dev, looks like its grabbed a bunch of the changes from my previous PR

NickPhura (approved):

PR #150, opened 2020-11-09
Context menu for main map and create point of interest

micheal-w-wells:
* Context menu on map on right click (currently just main map)
* Interactive Geos: parent component of map container can dictate colour, onclick callback, and popup content.
* Point of interest created when selected in context menu, on map right click


popkinj (approved):
Looks good. ☺

PR #154, opened 2020-11-12
62: single activity edit

NickPhura:
Some updates to the `/database` folder, specifically around migrations.

__Commit 1: the actual migration updates I'm making that relate to editing an activity record.__
- Remove the version column
- Add 2 new date columns (created and deleted timestamps)
- Add an index on the activity_id column
- Tweak some index names to match the default convention
- Merged the media_keys migration into the activity table migration

_I am technically being a bit naughty in updating the existing migration, but it was getting a bit ugly to throw some of the changes I wanted into their own migration.  And since we haven't gone to test or prod, can probably get away with it for now :).  I am planning to purge all of the bad dev data in Dev anyways, so this can line up with that._

__Commit 2: just a bunch of linting and some trivial updates to imports and dead-code.__
- Added the linter packages/files
- Ran the linter
- Added DB_SCHEMA in a bunch of places
- Removed some dead code


sdevalapurkar (commented):
looks pretty good to me, but will definitely let @micheal-w-wells sign off :)

micheal-w-wells:
Why ditch the version column?

NickPhura:
> 
> 
> Why ditch the version column?

I was trying to investigate ways to do versioning, and came across this post:
https://stackoverflow.com/questions/34571125/custom-serial-autoincrement-per-group-of-values

The poster in that thread wants a serial id for a group of values (which is what we were going for originally with the version column), and my take-away from the responses is that it is either more trouble than its worth, or inherently problematic with potential for race conditions.

The other thing is that the `received_timestamp` already tells us what the order of records is.  You could just do a `max(recieved_timestamp)` and that is the same result as `max(version)`.  Or more ideally, I am planning to have the update endpoint update the old records' `deleted_timestamp` and then for GETs we just have to do a  `select * from blah where activity_id = 12 and deleted_timestamp = null` which would be super fast.

micheal-w-wells:
Where I see the version column having value is when I'll inevitably get asked to make a table/view to show admin changes on certain types of records, giving the user a version number to refer to.  I'm game for the approach you outlined but maybe we can meet in the middle and add a computed version column based on timestamp.

micheal-w-wells (approved):

PR #160, opened 2020-11-17
62: Edit single Activity + Misc Updates

NickPhura:
62: Edit single Activity + Misc Updates
- Update form container logic
  - Specifically, there is a bug with the current RJSF package where error state isn't reset properly when the `submit` hook of the RJSF form leads to a re-render of the form (as it does with the search activity page). Work around for now is to manually reset some of that state during the `submit` workflow.
    - This isn't an issue with the regular `current activity` page, because we save the changed form data in the local database (which doesn't trigger a re-render of the form component).
- Update photo container (to take photo state props similar to the map and form containers)
- Add support for new created_timestamp activity column
- Fix a few instances of `disabled` that should have been `isDisabled`
- Other misc updates

### DIFF
--- a/app/src/components/activities-list/ActivitiesList.tsx
+++ b/app/src/components/activities-list/ActivitiesList.tsx
@@ -1,7 +1,6 @@
 import {
   Box,
   Button,
-  ButtonGroup,
   Checkbox,
   Divider,
   Grid,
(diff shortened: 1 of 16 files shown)


micheal-w-wells (approved):

sdevalapurkar (approved):
LGTM 💯

PR #167, opened 2020-11-17
964: Migrated lucy-web/api-mobile to invasivesbc/api

NickPhura:
- Wasn't planning to do this just yet, but I realized after merging my other PRs that they would be a bit messed up without this transition (because api-mobile had code that relies on the new code-tables, which are only in the new database in the new repo, etc).

- I ran it locally, and some basic operations seemed to be working as usual (specifically I tried the search/edit).

- The postman tests are likely broken simply because the current invasivesbc/api in dev is super out of date, so it wouldn't be working right now anyways.


micheal-w-wells:
I'm ok to merge it but we should do a de-stink commit soon

micheal-w-wells (approved):

NickPhura:
> 
> 
> I'm ok to merge it but we should do a de-stink commit soon

For sure, though skimming the smells, a bunch of them I'd argue are false-positives

PR #180, opened 2020-11-20
Fix state related race-condition on form save effects

NickPhura:
Some general improvements/fixes to the way the page loads and what happens when save actions occur.

### DIFF
--- a/app/src/features/home/search/SearchActivityPage.tsx
+++ b/app/src/features/home/search/SearchActivityPage.tsx
@@ -38,6 +38,8 @@ const SearchActivityPage: React.FC<ISearchActivityPage> = (props) => {
 
   const invasivesApi = useInvasivesApi();
 
+  const [isLoading, setIsLoading] = useState(true);
+
   const [geometry, setGeometry] = useState<Feature[]>([]);
   const [extent, setExtent] = useState(null);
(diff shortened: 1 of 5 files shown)


app/src/features/home/search/SearchActivityPage.tsx
```diff
@@ -125,7 +127,7 @@ const SearchActivityPage: React.FC<ISearchActivityPage> = (props) => {
    * @param {*} event the form change event
    */
   const onFormChange = useCallback(
-    debounced(500, (event: any) => {
+    debounced(100, (event: any) => {
```
sdevalapurkar, line 130: question:
why is this value lowered?

sdevalapurkar (dismissed):
lgtm

app/src/features/home/search/SearchActivityPage.tsx
NickPhura: I arbitrarily chose 500 to begin with, but found it was too long. 

Specifically, there was a funny interaction where if you changed the form, and then within 500 milliseconds updated the map, it would blip back to the old version of the form for a moment, because the map change would trigger a re-render before the 500 milliseconds delay expired. Likely wouldn't occur that much in real life, but lowering this number should fix that.

It is certainly something we can adjust as we play with it more.

micheal-w-wells (approved):

PR #181, opened 2020-11-20
134: Update S3 to work with gov object store config

NickPhura:
- Coincides with OpenShift secret updates for all 3 environments
  - secret: invasivesbci-object-store


sdevalapurkar (approved):
awesome, thank you!

micheal-w-wells (approved):

PR #182, opened 2020-11-24
Proper form content and behaviour for Creating a Terrestrial Plant Observation

### LINKED ISSUE #155: Forms: Create Terrestrial Plant Observation
**Describe the Task**
Ensuring the form content and behaviour is accurate for Creating a Terrestrial Plant Observation.

**Acceptance Criteria**
- [x] Activity status should be removed, this should not be editable or viewable as a dropdown to the user. 
- [x] Remove Activity Status from Postman test
- [x] Photo toggle should be removed from form, photo option available at bottom in photo section.
- [x] "Created date on device"/"Updated date on device" do not need to be visible in form. 
- [x] "Paper file" should say "Paper File ID"
- [x] Observation time & date should be auto-filled, but should be editable
- [ ] Time should reflect a 24 hr clock, not AM/PM.
- [x] "Type" should be removed as this reflects type of geometry that the user has selected, and this should  be auto-filled and not visible to the user in the front end. 
- [x] "Area" should be auto-filled based on the geometry the user has selected, and should be visible to the user but NOT editable. 
- [x] In image below, fields in blue should be auto-filled and not editable by user.
- [x] Remove the Sample ID variable as this is duplicated by the Sample Number collected with the Sample taken dialog
- [x] Set defaults for all radio buttons to "NO" when form loads
- [ ] Order of fields needs to be addressed once we have confirmation of how the form will be laid out, confirm with Mike Shasko.

sdevalapurkar:
## Overview

This PR has to do with the form content and behaviour for terrestrial plant observations. Some of the changes are explained below.

### Detailed Changes

1. It removes fields from the form such as:
- activity status
- photo toggle
- created and updated date on device
- type
- sample id

2. The observation date and time fields have been combined into one field called `observation_date_time` and this field is auto-populated with the current datetime if there is no existing value.

1. Area is auto-populated based on the geometry. Library https://www.npmjs.com/package/@turf/area was added to enable this, although manual calculation is done for circular geometries.

1. Paper file -> Paper File ID.

1. Auto-populate radio selects to `No`.

1. Update postman collection as required.

### Screenshots

Screenshot 2020-11-24 145457
Screenshot 2020-11-24 145522

### Issue Reference

https://github.com/bcgov/invasivesbc/issues/155


NickPhura:
Do you know what the units for the area are? Square meters?

Can we update the title of the area field to include the unit? 

`Area (m)`
Not sure if we can do superscript, otherwise `Area (m2)` with the 2 as a power, would be nice.

Unless Shasko thinks its not needed because they always use square meters so no one would be confused by what the number means?

sdevalapurkar:
Yup it is square meters (m^2)
I was thinking about that but wasn't sure if needed, I do think it is useful to display. Can definitely add it in :)

app/src/features/home/activity/ActivityPage.tsx
```diff
@@ -54,14 +56,68 @@ const ActivityPage: React.FC<IActivityPageProps> = (props) => {
...
 
+  /**
+   * Calculate the net area for the total geometry
+   * 
+   * @param {Feature} geoJSON The geometry in GeoJSON format 
+   */
+  const calculateGeometryArea = (geometry: Feature []) => {
+    let totalArea = 0;
+
+    if (!geometry || geometry.length === 0) return totalArea.toFixed(2);
```
NickPhura: Probably a subjective thing, but I'm not a big fan of single line calls like this.  Not sure why, maybe just because its not a convention seen very often, so it stands out a bit.

Also, trivial improvement, you can just say `if (!geometry || !geometry.length)` to get the same result

app/src/features/home/activity/ActivityPage.tsx
```diff
@@ -54,14 +56,68 @@ const ActivityPage: React.FC<IActivityPageProps> = (props) => {
 
   const [photos, setPhotos] = useState<IPhoto[]>([]);
 
+  /**
+   * Calculate the net area for the total geometry
+   * 
+   * @param {Feature} geoJSON The geometry in GeoJSON format 
+   */
+  const calculateGeometryArea = (geometry: Feature []) => {
```
NickPhura: Mega nits:
Technically the @param should describe an array of features: `Feature[]` 
`Feature []` has a space, I'm guessing typescript doesnt actually care tho?

app/src/features/home/activity/ActivityPage.tsx
```diff
@@ -54,14 +56,68 @@ const ActivityPage: React.FC<IActivityPageProps> = (props) => {
...
+   * 
+   * @param {*} doc The doc/activity object
+   */
+  const getDefaultFormDataValues = (doc: any) => {
+    const { activity_type_data } = doc.formData || {};
+
+    const areaOfGeometry = calculateGeometryArea(doc.geometry);
+    const observationDateTime = activity_type_data && activity_type_data.observation_date_time
+      ? activity_type_data.observation_date_time
+      : moment(new Date()).format();
```
NickPhura: A shorter way to do this is:
```
    const observationDateTime = activity_type_data && activity_type_data.observation_date_time || moment(new Date()).format();
```

NickPhura: It returns the last piece of the first non-falsy expression, so in this case, if `activity_type_data && activity_type_data.observation_date_time` is truthy, it returns `activity_type_data.observation_date_time`

NickPhura: Its basically a fancier ternary, because you can do any number of expressions in a chain, and have it pick whichever evaluates truthy first.

NickPhura (dismissed):
Looks great.  I had a couple of nit comments, but feel free to address them or not, they aren't hard issues by any means.

app/src/features/home/activity/ActivityPage.tsx
sdevalapurkar: for sure makes sense, good call!

NickPhura (approved):

PR #185, opened 2020-11-24
v1 iapp migrator

micheal-w-wells:
CLI tool to upload IAPP data (or other data provided it comes with a SiteID).

* Get's own token (doesn't yet prompt for creds)
* Creates one point of interest per site
* Filters for appropriate survey, treatment, and soon monitoring records by siteID and includes in point of interest data.
* TODO: prompt for creds, photos

from database directory:

tsc src/tools/IAPP_Migrator.tsx
node src/tools/IAPP_Migrator.js --help

Must include a site csv
other csvs optional.

### DIFF
--- a/database/package.json
+++ b/database/package.json
@@ -20,12 +20,17 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "dependencies": {
-    "@types/csv-parse": "^1.2.2",
+    "axios": "^0.21.0",
+    "csv-parser": "^2.3.3",
+    "csvtojson": "^2.0.10",
     "knex": "^0.21.6",
+    "meow": "^8.0.0",
     "pg": "~8.3.0",
+    "qs": "^6.9.4",
     "typescript": "~3.9.7"
   },
   "devDependencies": {
+    "@types/csvtojson": "^1.1.5",
     "@types/node": "^12.12.24",
     "@types/pg": "~7.14.4",
     "@typescript-eslint/eslint-plugin": "~3.7.1",
@@ -33,8 +38,8 @@
     "eslint-config-prettier": "~6.11.0",
     "eslint-plugin-prettier": "~3.1.4",
     "eslint": "~7.5.0",
-    "prettier": "^2.1.2",
     "npm-run-all": "~4.1.5",
+    "prettier": "^2.1.2",
     "ts-node": "~8.10.2"
   }
 }
--- a/database/src/tools/IAPP_Migrator.ts
+++ b/database/src/tools/IAPP_Migrator.ts
@@ -1,0 +1,262 @@
+import fs from 'fs';
+import { AxiosRequestConfig } from 'axios';
+import csv from 'csvtojson';
+import axios from 'axios';
+import qs from 'qs';
+
+const meow = require('meow');
+
+const cli = meow(
+  /*
+    Usage
+      $ IAPP_Migrate [[OPTION] [FILENAME] [OPTION] [FILENAME]...] ENDPOINT
+
+    Options
+      --site fileName, -si fileName
+      --survey fileName, -su fileName
+      --mechanicalTreatment fileName, -mt fileName
+      --monitoring fileName, -s fileName
+      --dispersal fileName, -d fileName
+      --bioControlOutput fileName, -b fileName
+
+    Examples
+      Load just sites:
+      $ IAPP_Migrate --site sites.csv http://point_of_interest_endpoint/
+      Load more:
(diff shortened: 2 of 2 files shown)


database/src/tools/IAPP_Migrator.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+
+const main = async () => {
+  const IAPPData = await loadAllData();
+
+  IAPPData.siteData.forEach(async (record: any) => {
+    if (record.OFF_Longitude) {
+      try {
+        // load relevant data from other CSV's:
+        const surveys = await IAPPData.surveyData?.filter((survey) => (survey as any).SiteID == record.SiteID)
```
micheal-w-wells: @NickPhura If I use a big spreadsheet it finds nothing, if i use a small one it does so I've clearly got an async bug in this file.  Anything you see just eyeballing it?  Otherwise does what its supposed to and will be straight forward to add photos later.

micheal-w-wells: To be clear, its the finding the related sites it has issues with (with lots of data) on line 159.

NickPhura: Async/await doesn't work with a forEach.  The awaits inside of each of the forEach loops should work (though I always have to double check because I always forget), but the forEach itself wont wait, it will loop over every item as fast as possible regardless of the internal awaits.  So part of the issue might be just that too many requests are getting slammed in too quickly?

If you want the loop to actually process each request one by one in order, you have to use another loop. `for` or `for...of` work well.

NickPhura: Otherwise,  I dont see where any other async stuff is happening here. Is the `IAPPData.surveyData` actually an asynchronous thing? Or is it just an array of objects? The `await` on line 159 might not be needed (though it wont break anything either).

micheal-w-wells: vscode seems to think i need that async for the loop for the awaits in it to work : /.  threw the await on 159 because it was only returning results if there was a small number of them. 

regardless I"ll try re-jigging it as a for...of and see if i have more luck.

database/src/tools/IAPP_Migrator.tsx
```diff
@@ -0,0 +1,254 @@
...
+    let counter = 0;
+  for(let arecord of IAPPData.siteData)
+  {
+    let record = arecord as any;
+
+    if (record.OFF_Longitude) {
+      try {
+        // load relevant data from other CSV's:
+        console.log('Finding surveys for Site: ' + record.SiteID + '...')
+        const surveys = IAPPData.surveyData?.filter((survey) => (survey as any).SiteID == record.SiteID)
```
micheal-w-wells: @NickPhura this only works if the survey csv is small.  Any other ideas?  I'm stumped.

database/src/tools/IAPP_Migrator.tsx
NickPhura: Yeah the async/await is needed inside the loop if you have async stuff happening there, but the loop itself wont honour them and will loop over its array regardless of the internal awaits. The only reason I bring it up is because maybe if it chucks out a billion requests all at once, its too much for the server to handle (potentially)

database/src/tools/IAPP_Migrator.tsx
```diff
@@ -0,0 +1,254 @@
...
+	  --site fileName, -si fileName
+	  --survey fileName, -su fileName
+	  --treatment fileName, -t fileName
+	  --monitoring fileName, -s fileName
+	  --dispersal fileName, -d fileName
+	  --bioControlOutput fileName, -b fileName
+
+  Examples
+    Load just sites:
+	  $ IAPP_Migragte --site sites.csv http://point_of_interest_endpoint/
```
sdevalapurkar:
```suggestion
	  $ IAPP_Migrate --site sites.csv http://point_of_interest_endpoint/
```

database/src/tools/IAPP_Migrator.tsx
```diff
@@ -0,0 +1,254 @@
...
+    Load more:
+	  $ IAPP_Migragte -si sites.csv -su surveys.csv -t treatments.csv -m monitoring.csv http://point_of_interest_endpoint
```
sdevalapurkar:
```suggestion
	  $ IAPP_Migrate -si sites.csv -su surveys.csv -t treatments.csv -m monitoring.csv http://point_of_interest_endpoint
```

sdevalapurkar (approved):

PR #186, opened 2020-11-25
Update POST /activity response

NickPhura:
The response isn't strictly needed, but the previous value was not useful in any way, as it was just the postgres serial id for the row.

The new response is the activity_id, which is the relevant id we use to track a unique activity.  That being said, the activity_id is already required to call the POST endpoint in the first place, so the caller should already have it on hand. 

So, It might be just as valid to have no response (other than the 200 OK).  Thoughts?

### DIFF
--- a/api/src/queries/activity-queries.ts
+++ b/api/src/queries/activity-queries.ts
@@ -64,7 +64,7 @@ export const postActivitySQL = (activity: ActivityPostRequestBody): SQLStatement
   sqlStatement.append(SQL`
     )
     RETURNING
-      activity_incoming_data_id;
+      activity_id;
   `);
 
   return sqlStatement;


sdevalapurkar (approved):
this is on successful POST creation of an activity right? So I think returning the `activity_id` makes sense even if it is known in first place. But shouldn't it be a 201 response?

NickPhura:
> 
> 
> this is on successful POST creation of an activity right? So I think returning the `activity_id` makes sense even if it is known in first place. But shouldn't it be a 201 response?

It could be a 201, though a 200 isn't wrong either. I'd probably just leave it for now, and we can probably have a ticket for an overall review of our error handling and messages and stuff all at once.

PR #188, opened 2020-11-25
Consolidated the PR API Test into the deploy pipeline as it needs to …

rstens:
…wait until it is deployed.


sdevalapurkar (dismissed):
LGTM thanks :)

.github/workflows/deploy.yml
```diff
@@ -286,6 +286,43 @@ jobs:
...
+  # Smoke Test the API
+  PR-API-Test:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.merged == false && github.base_ref == 'dev' }}
+    needs:
+      - deployDatabase
+      - deployAPI
+    strategy:
+      matrix:
+        node-version: [12.x]
```
NickPhura, line 298: Should this be node 10.x like the others? (might not really matter in the end)

rstens: Just running newman, not running our api code.

NickPhura (dismissed):

PR #193, opened 2020-11-27
Handle Auth on Mobile Interface/Token Expiry

sdevalapurkar:
## Overview

This PR enables offline/mobile access of our application without requiring the need to be authenticated through keycloak. The frontend application is still protected using keycloak when being used on the web application, but under circumstances where the mobile app is being used (no network connectivity, etc) then the user can still have access to the app without having the need to authenticate using keycloak.

The backend calls will not work when not authed through keycloak on frontend however because the endpoint routes are still protected and guarded by keycloak.

As for the token refresh on expiry, this is already being handled by the Keycloak wrapper library we are using and our event handler. When the user's access token expires, a refresh token is automatically issued and the user's session is re-validated meaning that they do not notice any difference on their end. The bearer token for axios calls is also automatically updated so the user does not notice any issues with subsequent backend calls 😄 

## How Has This Been Tested

Locally on both web as well as mobile (android app emulator) interfaces. Web requires and mobile does not require fe auth.

### Web Interface

web

### Mobile Interface

mobile

## Ticket Reference

https://github.com/bcgov/lucy-web/issues/868

### DIFF
--- a/app/public/keycloak.json
+++ b/app/public/keycloak.json
@@ -1,6 +1,6 @@
 {
     "realm": "dfmlcg7z",
-    "auth-server-url": "https://dev.oidc.gov.bc.ca/auth/",
+    "auth-server-url": "https://test.oidc.gov.bc.ca/auth/",
     "ssl-required": "external",
     "resource": "invasives-bc",
     "public-client": true,
--- a/app/public/keycloak.json
+++ b/app/public/keycloak.json
@@ -1,6 +1,6 @@
 {
     "realm": "dfmlcg7z",
-    "auth-server-url": "https://test.oidc.gov.bc.ca/auth/",
+    "auth-server-url": "https://dev.oidc.gov.bc.ca/auth/",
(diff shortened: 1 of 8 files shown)


micheal-w-wells (approved):

NickPhura:
Looks good.

Also, sort of related, I created a new ticket to track what I think will be the next iteration on some of the work you did here (around offline/mobile support in general): https://github.com/bcgov/invasivesbc/issues/194

NickPhura (approved):

rstens:
For future PRs, can you make the PR name a bit more descriptive instead of just the number? This is particularly useful for SonarCloud when you review the list of prs.

rstens (approved):

sdevalapurkar:
Oh @rstens sure you mean the branch name right? Thanks for letting me know I will do that next time onwards 👍🏼

rstens:
@sdevalapurkar  Yes

PR #196, opened 2020-11-30
Map Popout Component from the Bottom

sdevalapurkar:
# Overview

- Make popout component for map enter from bottom of screen and fix coordinates being undefined

## Type of change

- [x] Refactoring

## How Has This Been Tested?

- Local dev environment

## Screenshots

bottompopout

## Checklist:

- [x] I have made corresponding changes to the documentation
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/src/features/home/map/MapPage.tsx
+++ b/app/src/features/home/map/MapPage.tsx
@@ -149,12 +149,6 @@ const MapPage: React.FC<IMapProps> = (props) => {
     setContextMenuState({ ...contextMenuState, isOpen: false });
   };
 
-  //testing
-  const handleContextMenuOpen = () => {
-    setContextMenuState({ ...contextMenuState, isOpen: true });
-  };
-  //end testing
-
   const changeContextMenu = (targetContextMenu: contextMenuType) => {};
 
   // todo: handle closing of popup, this does not work:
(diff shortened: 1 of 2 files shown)


dogcomplex:
Looks great!  I can't speak to the above automated tests, but the code and screenshot look about right for intended effect.  I think in the future we might want some kind of 2-stage popup, where the first click of a point just pops up a one-line summary/header of the most pertinent details, and then a click of a "V" (down arrow) button (or "More...") or some similar concept has it pop-up entirely (or alternatively, scroll the page down while lengthening the form area, that way we don't lose any map space if the user scrolls up).   But this is all speculative behavior as far as what works best UX wise.

app/src/features/home/map/MapPage.tsx
```diff
@@ -27,16 +27,16 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-    height: '100%',
-    width: '66.66%'
+    height: '60%',
+    width: '100%'
   },
   popOutGridItemExpanded: {
-    height: '100%',
-    width: '33.33%'
+    height: '40%',
+    width: '100%'
```
NickPhura, line 35: Not a real comment, just a useful feature of material-ui: It has a breakpoints concept, so you can have different css depending on what the screen size is: https://material-ui.com/customization/breakpoints/

So you could make the bottom popup take up more or less of the page depending on screen size.  (Not something that necessarily needs to be accounted for in this PR)

dogcomplex (approved):
Looks good.  We just catted through the failing check and looks like an unrelated issue todo in another PR

PR #217, opened 2020-12-01
Update steps and description for postman tests

sdevalapurkar:
# Overview

- Update steps and description for postman tests

## Type of change

- [x] Documentation

## How Has This Been Tested?

- Just documentation in Github

### DIFF
--- a/testing/integration/postman/README.md
+++ b/testing/integration/postman/README.md
@@ -8,21 +8,19 @@ How to run the PostMan Tests:
 - Change the following settings in Postman environment DEV:
 
 ```json
-		{
-			"key": "POSTMAN_PASSWORD",
-			"value": "<<Change to istest1 pwd>>",
-			"enabled": true
-    }
+{
+	"key": "POSTMAN_PASSWORD",
+	"value": "<Change to openshift postman pwd>",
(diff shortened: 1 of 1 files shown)


testing/integration/postman/README.md
```diff
@@ -1,38 +1,24 @@
...
-		},
-		{
-			"key": "prNumber",
-			"value": "<<change to PR#>>",
-			"enabled": true
-		},
-		{
-			"key": "postman_pw",
+			"key": "POSTMAN_PASSWORD",
 			"value": "<<Change to istest1 pwd>>",
```
NickPhura: This isn't technically true anymore.  istest1 has a new password (due to the old one expiring and needing to be re-set).

You can see the secret we made to track those passwords here: https://console.pathfinder.gov.bc.ca:8443/console/project/8ecbmv-tools/browse/secrets/invasivesbc-test-users

Technically it is true that its the same as the istest2 user (though we could also just add a line to that secret to track the postman password separately, and then dont have to make this comparison)

sdevalapurkar: awesome idea, done! :)

NickPhura (approved):

PR #231, opened 2020-12-02
New elevation column

popkinj:
## Loading of contextual data, including elevation.

This PR includes logic for filling the activity incoming table with contextual data that is not available in the field.
@NickPhura - There is bare SQL in the new component file. It's there for development purposes only. When I've completed the rest of the fields I'll move into separate query files that imported ala the rest of the project.

Included here is an additional migration file to bring in sync with the migrations data table.


api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
+import axios from 'axios';
+import { getDBConnection } from '../database/db';
+import { getLogger } from './logger';
+
+const defaultLog = getLogger('activity');
```
NickPhura:
```suggestion
const defaultLog = getLogger('context-queries');
```

api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
...
+    },{
+      tableName: 'WHSE_ADMIN_BOUNDARIES.TADM_MOT_DISTRICT_BNDRY_POLY',
+      targetAttribute: 'DISTRICT_NAME',
+      targetColumn: 'moti_districts'
+    }
+  ];
+
+  /* For each layer run an asynchronous request */
+  for (let layer of layers) {
+    const url = `${api}/context/databc/${layer.tableName}?lon=${x}&lat=${y}`;
```
NickPhura, line 53: Is this making an API call to this API, to itself? (If I am misreading this then ignore the below)

If so, we definitely don't want to do that.  We should just pull the logic from the other endpoint into some re-usable functions, and then both endpoints can leverage it directly.

api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
...
+  const y = geom.anchorPointY;
+  const api = `${req.protocol}://${req.get('host')}/api`
+  const config = {
+    headers: {
+      authorization: req.headers.authorization
+    }
+  }
+
+  /* For each layer run an asynchronous request */
+  const url = `${api}/context/elevation?lon=${x}&lat=${y}`;
```
NickPhura, line 172: Same comment as above.

api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
...
+      targetAttribute: 'utm',
+      targetColumn: 'utm_zone'
+    }
+  ];
+
+  /* For each layer run an asynchronous request */
+  for (let layer of layers) {
+    const url = `${api}/context/internal/${layer.targetAttribute}?lon=${x}&lat=${y}`;
+
+    axios.get(url,config)
```
NickPhura, line 122: Same comment as above.

api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
...
+      .then(async (response) => {
+        const attribute = response.data.target[layer.targetAttribute];
+        const column = layer.targetColumn;
+        const connection = await getDBConnection();
+        const sql = `
+          update activity_incoming_data
+          set (${column}) = ('${attribute}')
+          where activity_id = '${id}'
+        `;
```
NickPhura, line 64: This sql is duplicated a few times in this file.  Would be good to move it into its own function, and then it can be re-used.

NickPhura (changes requested):

api/src/utils/context-queries.ts
popkinj: Hey @NickPhura 
This may be an opinion thing. I've always considered it best practices for an API to consume it's own resources.

api/src/utils/context-queries.ts
popkinj: Good eye @NickPhura 
Yes there is a lot of duplication... I will most likely optimize things when I've completed all the data loaders. I'm just doing this PR a tad prematurely to get dev running again.

api/src/utils/context-queries.ts
NickPhura: I agree the logic shouldnt be duplicated, but if the code from the other endpoint is pulled into a common function, then both of these endpoints can leverage it the same way. So they would both be doing things the same way, with the same result.  But you wouldn't need all of the extra axios/authorization code, and it wouldn't tack on an extra http request, which is the slowest part of the operation by far.

NickPhura: It would also help resolve your concurrency issue, because instead of calling 3 distinct endpoints via HTTP which have no concept of one another, you can just set up a transaction and make 3 sql calls, and then commit it.

And if the sql is created by a function that both endpoints leverage, then there is no issue in that being different either.

Similar to what I did here: https://github.com/bcgov/invasivesbc/blob/dev/api/src/paths/activity.ts#L310-L321

NickPhura: These 2 posts articulates some of the reasoning better than mysef: https://softwareengineering.stackexchange.com/questions/345406/is-it-bad-design-to-internally-call-api-endpoints-from-within-the-api-instance
https://www.reddit.com/r/django/comments/7sxiqn/internal_api_requests_does_it_make_sense_for_a/

NickPhura: @micheal-w-wells probably on you to decide ultimately which way we go with this kind of thing, as you'll be maintaining it longer-term.

api/src/utils/context-queries.ts
NickPhura: Gotcha, no worries then

api/src/utils/context-queries.ts
popkinj: Yeah.... those are good points... again. I think this is an opinion thing.
Having our own services for collecting data allows for other services to use them. It also allows for more flexibility when it comes to dataBC services that are protected... or in a format that isn't appropriate. Also the coordinate transforms can't directly access the database.

api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
...
+import { getDBConnection } from '../database/db';
+import { getLogger } from './logger';
+
+const defaultLog = getLogger('activity');
+
+/**
+ * Insert contextual data for the new activity record from
+ * the BC Geographic Warehouse (BCGW)
+ *
+ * @param id {integar} The record ID for the activity recently
```
sdevalapurkar:
```suggestion
 * @param id {integer} The record ID for the activity recently
```

api/src/utils/context-queries.ts
```diff
@@ -0,0 +1,204 @@
...
+      defaultLog.debug({ label: 'addingContext', message: 'error', error });
+    });
+}
+
+export const commit = function (record:any,req:any) {
+  const id = record.activity_id;
+  saveBCGW(id,req); // Insert DataBC BCGW attributes
+  saveInternal(id,req); // Insert local attributes
+  saveElevation(id,req); // Insert elevation
+};
```
sdevalapurkar, line 204:
```suggestion
};

```

api/src/paths/activity.ts
```diff
@@ -261,6 +262,9 @@ function createActivity(): RequestHandler {
 
       const result = (createResponse && createResponse.rows && createResponse.rows[0]) || null;
 
+      // kick off asynchronous context collection activities
+      commitContext(result,req);
```
NickPhura, line 210: This could be moved into the transaction above, and then the existing `connection` could be passed into it as a param.  And then if the `commit` function executes some SQL, it will all be part of this same transaction and using the same connection, so all of the gets and updates will all be atomic.

popkinj: That's a good idea. It make take care of the lock errors as well.

api/src/utils/context-queries.ts
NickPhura: I'm not sure I follow. I'm not suggesting that you get rid of the endpoints you've created, if they are useful for others.

All I'm saying is that this `commit` function may as well call the elevation/databc/etc endpoint code directly, rather than making an HTTP call to itself.

What I mean, as an example, is that the `/context/internal/{target}.ts` endpoint really just needs to wrap some logic in a bit of request/response handling.  But the core logic that matters could live in a re-usable function outside of that file.  And then the `/context/internal/{target}.ts` endpoint an just call it, and simply wrap the return value in a `res(200)` type call, as appropriate.  And then the `commit` function can also call it, but internally without needing all of the http request logic.

api/src/paths/activity.ts
NickPhura: This also needs to be added to the `updateActivity` function below as well, as the update is really performing a create behind the scenes, so it would also need these values added again.

api/src/utils/context-queries.ts
popkinj: Oh right. I see what you're saying. Yeah... that's definitely doable.

api/src/paths/activity.ts
popkinj: Nice catch... done.

api/src/utils/context-queries.ts
popkinj: haha... a copied spelling mistake. Fixed.

api/src/utils/context-queries.ts
popkinj: Future enhancement.

api/src/utils/context-queries.ts
popkinj: Future enhancement.

api/src/utils/context-queries.ts
popkinj: Done

NickPhura (approved):

PR #236, opened 2020-12-02
Refactor to reduce duplication - Part 3

sdevalapurkar:
# Overview

- Refactor uploadMedia function into shared location

## Type of change

## How Has This Been Tested?

Need to see on sonarcloud

### DIFF
--- a/api/src/paths/media.ts
+++ b/api/src/paths/media.ts
@@ -107,7 +107,7 @@ function getMedia(): RequestHandler {
  *
  * @returns {RequestHandler}
  */
-export function uploadMedia(label: String): RequestHandler {
+export function uploadMedia(label: string): RequestHandler {
   return async (req, res, next) => {
     defaultLog.debug({ label, message: 'uploadMedia', body: req.body });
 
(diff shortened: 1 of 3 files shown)


api/src/paths/media.ts
```diff
@@ -94,3 +97,60 @@ function getMedia(): RequestHandler {
...
+/**
+ * Uploads any media in the request to S3, adding their keys to the request, and calling next().
+ *
+ * Does nothing if no media is present in the request.
+ *
+ * TODO: make media handling an extension that can be added to different endpoints/record types
+ *
+ * @returns {RequestHandler}
+ */
+export function uploadMedia(label: string): RequestHandler {
```
NickPhura: I wouldn't worry about passing in a label.

Id just keep the label as 'uploadMedia' like on line 132.

sdevalapurkar: oh okay cool, I saw it was a custom label depending on where it was being used so I added that - can remove :)

NickPhura (approved):

api/src/paths/media.ts
NickPhura: Its really not a big deal either way, I think the cleanliness of the function (having no params) is probably just nicer than the extra label. especially as during dev/debugging, the flow of logs should be legible enough to know who called the uploadMedia function

PR #237, opened 2020-12-02
Sonarcloud: Fix a few code smells

sdevalapurkar:
# Overview

- Fix a few code smells

## Type of change

## How Has This Been Tested?

No real changes, mainly just clean up

### DIFF
--- a/api/src/paths/context/transform.ts
+++ b/api/src/paths/context/transform.ts
@@ -1,6 +1,6 @@
 'use strict';
 
-import { RequestHandler, Response } from 'express';
+import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { ALL_ROLES } from '../../constants/misc';
 import { getLogger } from '../../utils/logger';
--- a/app/src/components/status/StatusContainer.tsx
+++ b/app/src/components/status/StatusContainer.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { useInvasivesApi } from 'hooks/useInvasivesApi';
 
 const StatusContainer: React.FC = (props) => {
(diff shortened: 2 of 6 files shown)


app/src/components/status/StatusContainer.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { useInvasivesApi } from 'hooks/useInvasivesApi';
 
 const StatusContainer: React.FC = (props) => {
-  const invasivesApi = useInvasivesApi();
+  useInvasivesApi();
```
NickPhura: Looks like this isnt used at all.  May as well delete, and the import.

NickPhura (dismissed):

NickPhura (approved):

api/src/paths/context/transform.ts
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-import { RequestHandler, Response } from 'express';
```
micheal-w-wells, line 3: We should set up husky or something to strip unused imports on a pre or maybe even post commit hook, feels like I'm always cleaning these up behind myself.

micheal-w-wells (approved):

api/src/paths/context/transform.ts
NickPhura: One thing we should get into the habit of as an easy first step, which would need the whole team doing it to avoid merge issues, is running `lint fix` before putting up a PR.  It wont auto-fix ALL issues, but it will take care of most of the common ones.

APP: `npm run lint:fix`
API: `npm run lint-fix`

(The difference in commands could also be aligned, was a carry over from new vs old repo)

Issue #222, opened 2020-12-01
Geometry/Area size limit


ZoeSimon:
**Describe the Task**
There should be an error message provided to the user when they're creating a geometry.

- Terr Plant obs / treat 1 to 10000 sq m
- Aquatic plant obs/treat 1 to 100,000 sq m
- All animal activity 1 to 100,000

sdevalapurkar:
setup for such custom validation is taken care of in #241 
future area restrictions, etc can be added in matter of few lines in `customValidation.ts` file

PR #241, opened 2020-12-04
🌷 207: Finalize terrestrial plant form

### LINKED ISSUE #207: Forms: Finalize Terrestrial Plant Observation
**Describe the Task**
This is finalized workflow design/concept for the Terrestrial Plant Observation.

**Acceptance criteria**

- [x] Review the workflow overview on the image attached (Workflows_TerrPlantObs_v1.jpg)
- [x] first acquire the geometry. Note the 10,000sq m size limit (addressed in ticket #222 )
- [x] autofill the top elements of the form
- [x] keep the form a 3 columns and as condensed as possible
- [x] The dropdown and booleans MUST be functional and complete
- [x] The paper file ID inputs I believe are a "add" as required behaviour (not the static 3 columns I have depicted)
- [x] All booleans default to "no"
- [x] All fields are mandatory except the sample taken/sample number
- [ ] create the appropriate table in postgres (addressed in ticket #220 )
- [ ] load all content into the invasivesBC_observation table
- [ ] move to PO Review pipeline before closing / make changes accordingly

sdevalapurkar:
# Overview

- Finalizing plant terrestrial form
- Add missing elements to form (lat/lng), move elements to appropriate area of the form
- Custom validation for area limitations for different activity types (currently only validating plant terrestrial, but setup is there for adding that logic in the different activity types)

- https://github.com/bcgov/invasivesbc/issues/207
- https://github.com/bcgov/invasivesbc/issues/222

## Type of change

## How Has This Been Tested?

- Locally

## Screenshots

Screenshot 2020-12-04 092411


app/src/components/map/MapContainer.tsx
```diff
@@ -173,6 +173,8 @@ const MapContainer: React.FC<IMapContainerProps> = (props) => {
 
       if (feature.layerType === 'circle') {
         aGeo = { ...aGeo, properties: { ...aGeo.properties, radius: feature.layer.getRadius() } };
+      } else if (feature.layerType === 'rectangle') {
+        aGeo = { ...aGeo, properties: { ...aGeo.properties, isRectangle: true } };
```
NickPhura, line 177: Whats the need for tracking if its a rectangle?
I though that was a basic geojson shape already?

NickPhura: I see its referenced in the `calculateLatLng` function, can we not just use the polygon math to get its lat long?

I'm not totally familiar with the ways to get lat longs from a shape, so I'm probably just missing something.

NickPhura: To be clear, I'm just curious, I don't think its necessarily bad to track that its a rectangle, etc.

sdevalapurkar: great question :) the reason I needed to track whether the polygon is a rectangle or not is because if it is a reactangle we want the lat/lng to be the center of the polygon - but if it is a non-rect polygon, we want the lat/lng to be the first point used to draw it (anchor point)

app/src/rjsf/business-rules/customValidation.ts
```diff
@@ -0,0 +1,24 @@
+import { FormValidation } from '@rjsf/core';
```
NickPhura: This whole file: looks pretty solid, I've got some possible suggestions on how to improve it a bit (though I havent tested anything, so we'll see :))

sdevalapurkar: would love to hear and implement your suggestions!

NickPhura: Ping me when you're free

sdevalapurkar: thanks for the suggestion @NickPhura 😄

NickPhura (approved):

PR #243, opened 2020-12-04
Fix broken postman tests/collection

sdevalapurkar:
# Overview

- The `context_queries.ts` file that was added yesterday enforced a new property to be present on the body of requests and this was not present in our postman requests -> causing it to crash silently and result in a 500 instead of 200.

This PR fixes that issue.

## Type of change

## How Has This Been Tested?

- Locally using postman and collection

## Screenshots

Screenshot 2020-12-04 132726


popkinj (approved):
Looks good.

sdevalapurkar:
fixing this a different way, closing this for now

PR #248, opened 2020-12-07
🍒 654: Plant Chemical Treatment Form

sdevalapurkar:
# Overview

- https://github.com/bcgov/lucy-web/issues/654
- Finalize chemical plant treatment form based on Shasko needs
- Add custom validation for wind direction/speed relations

## Type of change

## How Has This Been Tested?

- Local
- Confirmed behaviour/look and feel with Shasko

## Screenshots

chem

### DIFF
--- a/api/src/openapi/api-doc.json
+++ b/api/src/openapi/api-doc.json
@@ -203,9 +203,6 @@
           "activity_data": {
             "$ref": "#/components/schemas/Activity"
           },
-          "activity_type_data": {
-            "$ref": "#/components/schemas/Treatment"
-          },
           "activity_subtype_data": {
             "$ref": "#/components/schemas/Treatment_ChemicalPlant"
           }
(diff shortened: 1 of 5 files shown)


api/src/openapi/api-doc.json
```diff
@@ -203,9 +203,6 @@
           "activity_data": {
             "$ref": "#/components/schemas/Activity"
           },
-          "activity_type_data": {
```
micheal-w-wells, line 206: Why are we axing this bucket?

sdevalapurkar: this is because for chemical plant, it doesn't have anything in the `activity_type_data`, everything is so custom that all its fields exist in the `subtype_data`. So in order to make the form look and behave as expected, this was removed

micheal-w-wells (approved):

PR #254, opened 2020-12-10
🧪 255: Chemical Treatment: Herbicide Autopopulate Application Rate Field

### LINKED ISSUE #255: Herbicide application rate autopopulate based on code for chemical treatments
**Is your feature request related to a problem? Please describe.**
When users select a liquid herbicide, there is a preset application rate for each herbicide and we want that to prepopulate in the field based on selection of a herbicide

**Describe the solution you'd like**
Herbicide application rate autopopulate based on code for chemical treatments
It should be editable and not enforced, but should autopopulate and only change when herbicide code changes (not on change of other form fields)

sdevalapurkar:
# Overview

- Chemical Treatment: Herbicide Autopopulate Application Rate Field
- Autopopulate herbicide application rate field, based on `liquid_herbicide_code` value
- Allow ability to edit application rate even after it has been auto populated
- Don't modify application rate field unless liquid herbicide code field changes
- #255 

## Type of change

## How Has This Been Tested?

- Locally, all use cases gone through

## Screenshots

Screenshot 2020-12-09 165601


app/src/constants/herbicides.ts
```diff
@@ -0,0 +1,40 @@
+export const Herbicides = {
```
NickPhura: Can you add some doc to this explaining what the values are, and why this exists in general.

Also, I would maybe suggest we move this into the business-rules section, since its very specific to the custom rules, and not so much to the general app.

app/src/rjsf/business-rules/populateCalculatedFields.ts
```diff
@@ -0,0 +1,62 @@
+import * as diff from 'fast-array-diff';
+
+import { Herbicides } from 'constants/herbicides';
+
+export function populateHerbicideRates(oldSubtypeData: any, newSubtypeData: any): any {
```
NickPhura, line 9: I would add some doc to this function explaining what this is doing.  Could include the fact that it sets a default rate, but only triggers if the herbicide changes, and not if the user manually edited the rate, etc.

NickPhura: In general I would add some extra comments throughout this code, just because its complicated enough that its worth documenting for the next person.

app/src/constants/herbicides.ts
sdevalapurkar: done ✔️

app/src/rjsf/business-rules/populateCalculatedFields.ts
sdevalapurkar: for sure sounds good!

app/src/features/home/search/SearchActivityPage.tsx
```diff
@@ -122,7 +122,7 @@ const SearchActivityPage: React.FC<ISearchActivityPage> = (props) => {
   /**
    * Save the form whenever it changes.
    *
-   * Note: debouncing will prevent this from running more than once per `500` milliseconds.
```
NickPhura, line 125: (Unrelated to this line)

I think we also need to add all of the custom validation and auto-populate logic to this component as well?

This is the page where an admin can edit/update an existing record.  I dont know what the likelihood of them changing the herbicide or the geometry are (I assume pretty low), but the area and rate fields, etc, should probably also update as they do in the 'regular' activity page. Thoughts?

This could be its own ticket/pr (as you see fit), depending on how much work you think it would be.  Hopefully its just a matter of calling the same functions here, too.

sdevalapurkar: Yes I think it would be better as its own ticket/PR :)

sdevalapurkar (commented):
added comments and addressed feedback :)

NickPhura (approved):

PR #257, opened 2020-12-11
✂️ 256: Ability to copy paste form data between activities

### LINKED ISSUE #256: Ability to copy paste activity form data from one activity to another
**Is your feature request related to a problem? Please describe.**
If many activities are created where all/most fields are similar, user has to recreate from scratch

**Describe the solution you'd like**
Ability to copy paste activity form data from one activity to another to avoid having to recreate from scratch

**Describe alternatives you've considered**
Maybe some clone feature could be a thing?

sdevalapurkar:
# Overview

- ✂️ Ability to copy paste form data between activities
- Can copy paste entire formdata from a given activity into another activity (with some exceptions such as geo info) based on some business rules
- #256 

## Type of change

## How Has This Been Tested?

- Locally

## Screenshots

copypaste

### DIFF
--- a/app/src/components/form/FormContainer.tsx
+++ b/app/src/components/form/FormContainer.tsx
@@ -1,22 +1,26 @@
 import { Box, CircularProgress, ThemeProvider, Typography } from '@material-ui/core';
 import { IChangeEvent, ISubmitEvent } from '@rjsf/core';
 import Form from '@rjsf/material-ui';
+import { DatabaseContext } from 'contexts/DatabaseContext';
 import { ActivitySyncStatus } from 'constants/activities';
 import { useInvasivesApi } from 'hooks/useInvasivesApi';
 import { JSONSchema7 } from 'json-schema';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useContext } from 'react';
 import ArrayFieldTemplate from 'rjsf/templates/ArrayFieldTemplate';
 import FieldTemplate from 'rjsf/templates/FieldTemplate';
 import ObjectFieldTemplate from 'rjsf/templates/ObjectFieldTemplate';
 import RootUISchemas from 'rjsf/uiSchema/RootUISchemas';
 import rjsfTheme from 'themes/rjsfTheme';
 import FormControlsComponent, { IFormControlsComponentProps } from './FormControlsComponent';
+import { notifySuccess } from 'utils/NotificationUtils';
+import { saveFormDataToSession } from 'utils/saveRetrieveFormData';
 
 export interface IFormContainerProps extends IFormControlsComponentProps {
   classes?: any;
   activity: any;
   customValidation?: any;
   isDisabled?: boolean;
+  pasteFormData?: Function;
   /**
    * A function executed everytime the form changes.
    *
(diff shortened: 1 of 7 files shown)


app/src/components/form/FormContainer.tsx
```diff
@@ -59,12 +65,25 @@ const FormContainer: React.FC<IFormContainerProps> = (props) => {
     return <CircularProgress />;
   }
 
+  const copyFormData = () => {
```
NickPhura: Could move this into the ActivityPage component, and pass it in as a prop.
Mainly just because it would align with the current style of all functions being controlled by the parent, and passed in.

app/src/rjsf/business-rules/formDataCopyFields.ts
```diff
@@ -0,0 +1,12 @@
+/*
+  Function to determine which fields to copy from provided activity_data/activity_type_data/activity_subtype_data
+  and apply some filters based on the type of data 
+*/
+export function getFieldsToCopy(data: any, type: string) {
+  if (type === 'Activity') {
+    const { activity_date_time, latitude, longitude, reported_area, ...activityDataToCopy } = data;
```
NickPhura, line 12: Maybe include a comment here that explains that these fields are not being included because they are read-only fields that are generated based on other information which can't be copied.

Actually, now that I say that, I wonder if Shasko will want the geometry to also be copied? If he wants the copy button to really be a copy activity button, rather than a copy form button?  An enhancement for later perhaps.

NickPhura: Also, not a real comment, but just as an alternative method:

You can also use `delete` to remove properties from an object.

``` javascript
delete data.activity_date_time;
delete data.latitude;
// ...etc

return data;
```

app/src/utils/saveRetrieveFormData.ts
```diff
@@ -0,0 +1,36 @@
...
+  (due to business rule)
+*/
+export function retrieveFormDataFromSession(activity: any): any {
+  const formData = activity.formData;
+  const activityData = { ...formData.activity_data };
+  const newActivityData = { ...activityData, ...JSON.parse(sessionStorage.getItem('copiedFormData')).activity_data };
+
+  return {
+    ...JSON.parse(sessionStorage.getItem('copiedFormData')),
+    activity_data: newActivityData
```
NickPhura, line 34: Why do we need to combine the `activity_data` (lines 27-30) separately from the rest of the form data?

(excluding null checks) could we not just say:

```javascript
export function retrieveFormDataFromSession(activity: any): any {
  return {
    ...JSON.parse(sessionStorage.getItem('copiedFormData')),
    ...activity.formData
 };
}
```

app/src/components/form/FormContainer.tsx
sdevalapurkar: sounds good!

app/src/utils/saveRetrieveFormData.ts
sdevalapurkar: this will just copy whatever is in `activity.formData` I believe based on my testing
I need to combine activity_data separately so that the fields that are copied are pasted but the existing lat/lng/area/datetime are not overriden

NickPhura (approved):
Nice one

PR #258, opened 2020-12-11
✅ ❌ 239 & 246: Custom error and success messages for sync activities

sdevalapurkar:
# Overview

This PR includes the following proposed change(s):

- Adding custom error and success message popups for sync activity failing and working cases
- #239 
- #246 

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- Locally, tested functionality

## Screenshots

formerrror

PR #259, opened 2020-12-11
194: Add network plugin support

### LINKED ISSUE #194: Context for network and other device information
The Invasives app needs to track changes to the network connection so it can alter its behaviour as needed.

Capacitor has a built in network plugin: https://capacitorjs.com/docs/apis/network

Depending on how the plugin can be used, we need to device some way for the current network state to be seen by any most components in the app, so that they can grey out, or display user messaging, or fetch keycloak tokens, etc, as needed.

A good approach:
The plugin could be turned into a context, and then any component in the app could access it.
We could also include in this new context, any other related plugins or device info that the app might want to access.
So instead of a network context, it would be a generic context providing any number of ionic/capacitor/cordova information (as needed).
The value of the context could then be an object with various properties (one for network, one for battery life, one for any other device related details, etc).

NickPhura:
# Overview

- New context that provides network information
- Added a footer bar to display network information
  - This isn't necessarily the only place to show it, but is working first pass.
- Added titles to the tabs (they show up in the browser tab)
- Updated some `div` to `Box`, in keeping with material-ui style.
  - Moved some css from useStyles into the inline settings of the Box component
- Updated `initConfig` to `keycloakConfig` just for readability
- Misc updates/tweaks

## Type of change

- [ ] Version change

## How Has This Been Tested?

Locally.  The main change is the addition of the network context which works (and is simple).  
More testing will be needed to ensure the larger goal of moving between offline/online and syncing and keycloak tokens is working.  Possibly some discussions around how best to manage those states (online vs offline), and what considerations there are around the user workflow.

## Screenshots

image

## Checklist:

~~[ ] I have made corresponding changes to the documentation~~
~~[ ] New and existing unit tests pass locally with my changes~~

### DIFF
--- a/api/src/paths/activity/{activityId}.ts
+++ b/api/src/paths/activity/{activityId}.ts
@@ -124,7 +124,7 @@ function getMedia(): RequestHandler {
 
     const result: IMediaItem[] = response.map((s3Object: GetObjectOutput) => {
       // Encode image buffer as base64
-      const contentString = Buffer.from(s3Object.Body).toString('base64');
+      const contentString = Buffer.from(s3Object.Body as Buffer).toString('base64');
 
       // Append DATA Url string
       const encodedFile = `data:${s3Object.ContentType};base64,${contentString}`;
--- a/app/src/AppRouter.tsx
+++ b/app/src/AppRouter.tsx
@@ -1,47 +1,52 @@
+import { CircularProgress } from '@material-ui/core';
+import { NetworkContext } from 'contexts/NetworkContext';
 import HomeRouter from 'features/home/HomeRouter';
 import AuthLayout from 'layouts/AuthLayout';
 import PublicLayout from 'layouts/PublicLayout';
 import AccessDenied from 'pages/misc/AccessDenied';
 import { NotFoundPage } from 'pages/misc/NotFoundPage';
-import React from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Redirect, Switch } from 'react-router-dom';
 import AppRoute from 'utils/AppRoute';
 
 interface IAppRouterProps {
   deviceInfo: any;
   keycloak: any;
-  initConfig: any;
+  keycloakConfig: any;
 }
 
 const AppRouter: React.FC<IAppRouterProps> = (props) => {
-  const layout = props.deviceInfo ? PublicLayout : AuthLayout;
+  const networkContext = useContext(NetworkContext);
+
+  const [layout, setLayout] = useState<React.FC<any>>(null);
 
   const getTitle = (page: string) => {
     return `InvasivesBC - ${page}`;
   };
 
+  useEffect(() => {
+    // If on mobile and have no internet connection, then bypass keycloak
+    const layout = window['cordova'] && !networkContext?.connected ? PublicLayout : AuthLayout;
+
+    setLayout(() => layout);
+  }, [networkContext]);
+
+  if (!layout) {
+    return <CircularProgress />;
+  }
+
   return (
     <Switch>
       <Redirect exact from="/" to="/home" />
-      <AppRoute
-        path="/forbidden"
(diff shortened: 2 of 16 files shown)


app/src/components/network/NetworkStatusComponent.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React, { useContext, useEffect, useState } from 'react';
+
+interface INetworkStatusComponentProps {
+  classes?: any;
+}
+
+const NetworkStatusComponent: React.FC<INetworkStatusComponentProps> = (props) => {
+  const networkContext = useContext(NetworkContext);
+
+  const defaultConnectionStatusString = (!window['cordova'] && 'Online') || '';
```
NickPhura, line 12: The reason to default to 'Online' if we are on web, is that the network plugin is null (since its a mobile plugin).  So if we are on web, then just show 'Online' always, basically.  Future change is to maybe not show this information at all, since its not really relevant in a web setting.

NickPhura: This would be contingent on where this information gets shown in the end as well.  As the footer bar might not be where it ends up (depending on UX/PO feedbacK, etc)

app/src/components/tabs/TabsContainer.tsx
```diff
@@ -3,18 +3,11 @@ import { Assignment, Bookmarks, Explore, HomeWork, Map, Search } from '@material
 import React, { useCallback, useEffect } from 'react';
 import { useHistory } from 'react-router-dom';
 
-function a11yProps(index: any) {
```
NickPhura, line 6: These were defaults copied over from a sample app.  As we dont set these anywhere else, for now its just cruft.

app/src/features/home/HomeRouter.tsx
```diff
@@ -17,39 +17,75 @@ interface IHomeRouterProps {
 }
 
 const HomeRouter: React.FC<IHomeRouterProps> = (props) => {
+  const getTitle = (page: string) => {
```
NickPhura, line 20: Most of this is just formatting changes, due to the additional title prop making the lines too long.

api/src/paths/activity/{activityId}.ts
```diff
@@ -124,7 +124,7 @@ function getMedia(): RequestHandler {
 
     const result: IMediaItem[] = response.map((s3Object: GetObjectOutput) => {
       // Encode image buffer as base64
-      const contentString = Buffer.from(s3Object.Body).toString('base64');
```
NickPhura, line 127: No idea why this is suddenly an issue, but its the same reason the other open PRs API's failed to build (#257 #258)

app/src/AppRouter.tsx
```diff
@@ -1,47 +1,52 @@
...
 
   const getTitle = (page: string) => {
     return `InvasivesBC - ${page}`;
   };
 
+  useEffect(() => {
+    // If on mobile and have no internet connection, then bypass keycloak
+    const layout = window['cordova'] && !networkContext?.connected ? PublicLayout : AuthLayout;
+
+    setLayout(() => layout);
```
sdevalapurkar, line 31: why not just:
```suggestion
    setLayout(layout);
```

sdevalapurkar (dismissed):
@NickPhura this looks awesome, thanks! LGTM 🥇

sdevalapurkar (approved):

sdevalapurkar:
looks like there are some code smells and bit of duplication being reported by sonarcloud - would be nice to clean those up prior to merge if possible :)

but can address them at later point too!

app/src/AppRouter.tsx
NickPhura: if you dont wrap it, the component actually gets initialized during this call, which then fails because its not properly set up, etc.

PR #277, opened 2020-12-16
Wells 216

popkinj:
# Well Data
Querying and saving distance to closest well within the `activity_incoming_data` table. This happens automatically as data is synced with the database. If there are no wells within 500 meters... No value is inserted.

This is the first API Context end point that acts as an AJAX and Module target. The logic will be propagated to the other Context end points. In addition, the SQL formatting tool has been adapted to match other logic within the API. This too will be propagated to the other end points.

### DIFF
--- a/api/src/paths/context/well.ts
+++ b/api/src/paths/context/well.ts
@@ -76,7 +76,9 @@ GET.apiDoc = {
  *     lon: xxx.xxxxx,
  *     lat: xx.xxxxx
  *    }}
- * @return {RequestHandler}
+ * @param res {object} Express response object. If missing it is a module call
+ * @param next {function} Callback function. Gets passed the data from the API
+ * @return {object} The closest well object. Including a distance to the supplied location.
  */
 function getWell(req, res, next) {
 
--- a/api/src/paths/context/well.ts
+++ b/api/src/paths/context/well.ts
@@ -1,0 +1,101 @@
+'use strict';
+
+import axios from 'axios';
+import { RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { ALL_ROLES } from '../../constants/misc';
+import { getLogger } from '../../utils/logger';
+
+const defaultLog = getLogger('activity');
+
+
+export const GET: Operation = [getWell()];
+
+GET.apiDoc = {
+  description: 'Fetches the distance to the closest well in meters from a given location.',
+  tags: ['activity','databc'],
+  security: [
+    {
+      Bearer: ALL_ROLES
+    }
+  ],
+  parameters: [
+    {
+      in: 'query',
+      name: 'lon',
+      required: true
+    },
+    {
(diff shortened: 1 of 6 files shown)


api/src/paths/context/well.ts
```diff
@@ -0,0 +1,194 @@
...
+    .catch(failure);
+  
+}
+
+
+/* ## proxyWell
+  This allows us to export the getWell function for 
+  ECMAScript module usage.
+ */
+function proxyWell(): RequestHandler {
```
NickPhura, line 186: This is looking pretty good. 

The thing I would adjust a bit though, is that the `getWell` function shouldn't know its being called by an endpoint.  It should just receive the lat long (or an object containing the lat long), run some logic on it, and return the resulting bundle. And it doesnt know it was called by an endpoint, or that there is a request/response object, etc.

So instead of passing in `req`, `res`, and `next`, I would parse the request body out in the `proxyWell` function and pass those values or that object it to the `getWell` function directly.

The `getWell` function can then do all of its internal logic, and at the end just return a bundle or throw an error.

The `proxyWell` function can then either call `return res.status(201).json(bundle);` to return the bundle or return some http error if it caught an error.

This way, if someone else wants to call `getWell` internally, they don't have to mimic a request.  They just pass in the lat/long parameters it needs, and they get a bundle back, that they can do w/e they want with it.

popkinj: Yeah that's doable.

popkinj: Hmmmm.... wait a sec. There's functions inside getWells that needs the express res object. So don't think that will work.

NickPhura: I think that part of it just needs to be moved up a level into the proxyWell function

Like, the `moduleReturn` just needs to return the data.  Rather than calling `next(data)`.

NickPhura: All of the req/res/next stuff would come out of `getWell`, so anything that currently leverages that would need to adjust accordingly, and just return the results back to wellProxy, and let it handle calling next or w/e it needs to do

popkinj: Hey @NickPhura 
I believe I'll be extending this end point in the future.... As @mikeshasko will probably want more then one feature returned for the map layer. Maybe we can revisit then.

NickPhura: Up to you.

Though I'd argue it is less work to update it now, before a whole bunch of other places start calling it and then also need updating when this function changes down the road.

popkinj: Thanks @NickPhura 
I've already spent too much time on this. Will revisit if it becomes a problem.

NickPhura (approved):

popkinj:
Thanks @NickPhura .
Feel free to make any adjustments in the future. I'm not protective of my code ☺

Issue #266, opened 2020-12-15
Linking Monitoring to a Treatment


ZoeSimon:
**Describe the task**
In the references page, having the ability to list all treatments & create a monitoring based on a given treatment

**Acceptance Criteria**
- [x] User should be able to click on a given treatment and user should be taken to create monitoring page with information stored about that treatment ID

PR #280, opened 2020-12-17
🔗 266: Link Monitorings to Treatments

sdevalapurkar:
# Overview

- #266 
- Ability to have linked monitorings to a treatment (to create a monitoring you select a reference treatment activity and then create one based on it) 🔗 
- Can create many monitorings based off a single treatment
- The treatment id field will populate in the form for the monitoring

## Type of change

## How Has This Been Tested?

- Locally 

## Screenshots

sauce1
sauce2

### DIFF
--- a/app/src/constants/activities.ts
+++ b/app/src/constants/activities.ts
@@ -50,3 +50,21 @@ export enum FormValidationStatus {
   INVALID = 'Invalid',
   VALID = 'Valid'
 }
+
+export interface IActivity {
+  _id: string,
+  activityId: string,
+  docType: string,
+  activityType: ActivityType,
+  activitySubtype: ActivitySubtype,
+  status: string,
+  sync: {
+    ready: boolean,
+    status: string,
+    error: string
+  },
+  dateCreated: Date,
+  dateUpdated: Date,
+  formData: any,
+  formStatus: string
+}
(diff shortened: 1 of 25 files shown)


app/src/utils/addActivity.ts
```diff
@@ -0,0 +1,47 @@
...
+  const id = uuidv4();
+  const formData = !linkedRecordId
+    ? null
+    : {
+        activity_type_data: {
+          activity_id: linkedRecordId
+        }
+      };
+
+  const doc = {
```
NickPhura: (Not something that needs to be defined now, unless you want to)

One thing that Ive been meaning to do forever now, is create an actual interface for an activity.  Would help when combining/spreading objects to create a new activity, etc, as it will complain if required fields are missing, etc, etc.

sdevalapurkar: Added something, lmk if that's kinda what you had in mind :)

NickPhura (dismissed):

app/src/constants/activities.ts
```diff
@@ -50,3 +50,21 @@ export enum FormValidationStatus {
   INVALID = 'Invalid',
   VALID = 'Valid'
 }
+
+export interface IActivity {
```
NickPhura: Looks good, there is an interfaces folder though, that this could go in.
Maybe just make a new file called `activity-interfaces.ts` or something.

NickPhura (approved):

PR #283, opened 2020-12-17
✅ Fix sonar issues :)

sdevalapurkar:
## Overview

This PR aims to clean up some sonarcloud issues that are being reported. Not the entire list, but a good start hopefully! 🗑️ ✅

### DIFF
--- a/app/src/features/home/activity/ActivityPage.tsx
+++ b/app/src/features/home/activity/ActivityPage.tsx
@@ -63,16 +63,16 @@ const ActivityPage: React.FC<IActivityPageProps> = (props) => {
   /**
    * Set the default form data values
    * 
-   * @param {*} doc The doc/activity object
+   * @param {*} activity The doc/activity object
    */
-  const getDefaultFormDataValues = (doc: any) => {
-    const { activity_data } = doc.formData || {};
+  const getDefaultFormDataValues = (activity: any) => {
+    const { activity_data } = activity.formData || {};
 
-    const areaOfGeometry = calculateGeometryArea(doc.geometry);
+    const areaOfGeometry = calculateGeometryArea(activity.geometry);
     const activityDateTime = activity_data && activity_data.activity_date_time || moment(new Date()).format();
 
     return {
-      ...doc.formData,
+      ...activity.formData,
       activity_data: {
         ...activity_data,
         activity_date_time: activityDateTime,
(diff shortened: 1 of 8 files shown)


app/src/utils/PrivateRoute.tsx
```diff
@@ -20,9 +20,9 @@ const PrivateRoute: React.FC<IPrivateRouteProps> = (props) => {
   return (
     <Route
       {...rest}
-      render={(props) => (
+      render={(prop) => (
         <Layout>
-          <Component {...props} {...rest.componentProps} />
```
NickPhura, line 25: odd change?

sdevalapurkar: was complaining that props is defined above

app/src/features/home/activity/ActivityPage.tsx
```diff
@@ -123,11 +123,11 @@ const ActivityPage: React.FC<IActivityPageProps> = (props) => {
...
    *
-   * @param {IPhoto} photos An array of photo objects.
+   * @param {IPhoto} photosArr An array of photo objects.
    */
-  const savePhotos = async (photos: IPhoto[]) => {
+  const savePhotos = async (photosArr: IPhoto[]) => {
     await databaseContext.database.upsert(doc._id, (dbDoc) => {
-      return { ...dbDoc, photos: photos, dateUpdated: new Date() };
+      return { ...dbDoc, photos: photosArr, dateUpdated: new Date() };
     });
```
NickPhura, line 131: This is totally subjective (and I really don't care enough one way or the other), but id argue `photos` with the plural `s` is already indicative of an array.  Or at least, thats generally what my go-to is when naming things.  Also the type being an array should enforce its accessed appropriately, etc.

sdevalapurkar: yup! once again it was changed because of the same name being present in above scope issue

sdevalapurkar: otherwise, I also agree photos is a fine name

app/src/utils/PrivateRoute.tsx
NickPhura: oh gotcha

NickPhura (approved):

PR #289, opened 2020-12-21
Prettier linting fixes (FE, formatting)

sdevalapurkar:
# Overview

- Prettier linting fixes frontend
- 189 -> 29 errors

## Type of change

## How Has This Been Tested?

- Local, just formatting changes

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -10,7 +10,7 @@
   "requirePragma": false,
   "insertPragma": false,
   "proseWrap": "never",
-  "endOfLine": "lf",
+  "endOfLine": "auto",
   "arrowParens": "always",
   "htmlWhitespaceSensitivity": "ignore"
 }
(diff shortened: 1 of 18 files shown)


.prettierrc
```diff
@@ -10,7 +10,7 @@
   "requirePragma": false,
   "insertPragma": false,
   "proseWrap": "never",
-  "endOfLine": "lf",
+  "endOfLine": "auto",
```
NickPhura, line 13: Question: Whats the reasoning here?

sdevalapurkar: my machine had like 4500+ linting complaints about
```
[eslint] Delete 'cr' [prettier/prettier]
```
and it turns out that setting it to auto helps resolve issue of lf vs crlf

NickPhura: Gotcha.  Yeah probably has to do with what your local git is configured to do when checking out code (leave as LF vs convert to CRLF, etc).

In any case, it doesnt really matter what the setting is, as long as we are only committing LF line endings.  Possibly git is smart enough to take care of that one its own?  This only really came up because 1 file at one point had the wrong endings somehow, but I cant remember the context in which that happened.

NickPhura (approved):

PR #290, opened 2020-12-22
Frontend linting fixes and action

sdevalapurkar:
# Overview

- Frontend linting fixes and action
- Down to 0 errors, 14 warnings

## Type of change

- [x] Technical debt

## How Has This Been Tested?

- Everything still seems to be working locally

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
         with:
           node-version: "10"
       - name: Install dependencies
-        run: npm ci
+        run: npm install
         working-directory: app
       - name: Run linter
         run: CI=true npm run lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,6 +4,7 @@ on: [pull_request]
 
 jobs:
   build:
+    name: Lint App
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,19 @@
+name: Linting Checks
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: "10"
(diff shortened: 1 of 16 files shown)


dogcomplex:
Wanted to pass on the good will and review this - looking good, simple changes but make sense (so far).  Can follow up when out of draft status 👍

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
+name: Linting Checks
+
+on: [pull_request]
+
+jobs:
+  build:
```
NickPhura, line 6: Should add a name param below (will make the job title more readable in the actions view)

```yaml
  build:
    name: Lint App
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: "10"
+      - name: Install dependencies
+        run: npm install
+        working-directory: app
```
NickPhura, line 17: Can we add either a step, or a job, that also lints the api and database folders

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
+name: Linting Checks
```
NickPhura, line 1: I was actually playing around with a workflow for this yesterday, but never quite finished messing with it all.  So this is great!

.github/workflows/lint.yml
sdevalapurkar: I will do this later, as if I do it now the api and db lint will fail and we will have a failing action
I will work on lint fixes first then add the workflows to this respectively

.github/workflows/lint.yml
sdevalapurkar: yeah I would like this action to be a required check on pr merge too, that way we dont go backwards in terms of cleanly linted code :)

NickPhura: yeah for sure

.github/workflows/lint.yml
NickPhura: Ah fair point

NickPhura (approved):

PR #292, opened 2020-12-22
Lint API

sdevalapurkar:
# Overview

- Lint the API
- Down to just 3 warnings and 0 errors

## Type of change

## How Has This Been Tested?

- Local

### DIFF
--- a/api/src/paths/activity/{activityId}.ts
+++ b/api/src/paths/activity/{activityId}.ts
@@ -63,7 +63,7 @@ GET.apiDoc = {
  * @return {RequestHandler}
  */
 function getActivity(): RequestHandler {
-  return async (req, res, next) => {
+  return async (req, _, next) => {
     defaultLog.debug({ label: '{activityId}', message: 'getActivity', body: req.params });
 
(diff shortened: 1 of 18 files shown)


api/src/paths/activity/{activityId}.ts
```diff
@@ -63,7 +63,7 @@ GET.apiDoc = {
  * @return {RequestHandler}
  */
 function getActivity(): RequestHandler {
-  return async (req, res, next) => {
```
NickPhura: Hmmmm, is this the suggested course of action by the linter?

I feel like `res` is better in this case, as `_` is really just a different name for an unused parameter.  And this is more likely to result in someone actually using `_` to do stuff: `_.someAction()` type thing.

sdevalapurkar: nope, just used it cause it was unused
but I agree with you, will put it back :)

dogcomplex:
Scanned through and looking good 👍

dogcomplex (approved):
👍

PR #354, opened 2021-01-08
🧰 Setting up tooltips for form fields

sdevalapurkar:
# Overview

- Setting up ability to have tooltips for form fields

## Type of change

## How Has This Been Tested?

- Locally

## Screenshots

tooltip

### DIFF
--- a/app/src/rjsf/templates/ObjectFieldTemplate.tsx
+++ b/app/src/rjsf/templates/ObjectFieldTemplate.tsx
@@ -56,7 +56,11 @@ const ObjectFieldTemplate = (props: ObjectFieldTemplateProps) => {
             style={{ marginBottom: '10px' }}>
             <>
               {!invalidTooltipElements.includes(element.name) && (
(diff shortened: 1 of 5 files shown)


app/src/rjsf/templates/ObjectFieldTemplate.tsx
```diff
@@ -52,7 +54,17 @@ const ObjectFieldTemplate = (props: ObjectFieldTemplateProps) => {
             }
             key={index}
             style={{ marginBottom: '10px' }}>
-            {element.content}
+            <>
+              {!invalidTooltipElements.includes(element.name) && (
```
NickPhura: I believe we have access to the schema and uichema for the field being processed, so we should be able to just check if 'x-tooltip-text' exists, and use that to decide if we should add a tooltip or not.

NickPhura: It should be part of the `props: ObjectFieldTemplateProps`

NickPhura (approved):

PR #356, opened 2021-01-09
🔄 Ability to clone activities

### LINKED ISSUE #335: Ability to clone activity records
# Overview

Add ability to clone an activity as is. Will be useful in workflows as described in ticket #276 for example

sdevalapurkar:
# Overview

- Ability to clone activity records

## Type of change

## How Has This Been Tested?

- Locally

## Screenshots

Screenshot 2021-01-08 121202

### DIFF
--- a/app/src/components/activity/ActivityComponent.tsx
+++ b/app/src/components/activity/ActivityComponent.tsx
@@ -1,96 +1,43 @@
-import { Accordion, AccordionDetails, AccordionSummary, Typography, Button, Box } from '@material-ui/core';
-import { ExpandMore, FileCopy } from '@material-ui/icons';
-import { DocType } from 'constants/database';
-import { DatabaseContext } from 'contexts/DatabaseContext';
+import { Accordion, AccordionDetails, AccordionSummary, Typography } from '@material-ui/core';
+import { ExpandMore } from '@material-ui/icons';
 import FormContainer, { IFormContainerProps } from 'components/form/FormContainer';
 import MapContainer, { IMapContainerProps } from 'components/map/MapContainer';
 import PhotoContainer, { IPhotoContainerProps } from 'components/photo/PhotoContainer';
-import React, { useContext, useEffect, useState } from 'react';
-import { useHistory } from 'react-router-dom';
-import { addActivityToDB } from 'utils/addActivity';
-import { notifySuccess } from 'utils/NotificationUtils';
+import React from 'react';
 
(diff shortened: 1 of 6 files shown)


app/src/components/activity/ActivityComponent.tsx
```diff
@@ -1,96 +1,43 @@
...
         <>
           <Accordion>
             <AccordionSummary expandIcon={<ExpandMore />} aria-controls="panel-form-content" id="panel-form-header">
-              <Typography className={linkedActivityProps.classes.heading}>Linked Activity Form</Typography>
+              <Typography className={props.classes.heading}>Linked Activity Form</Typography>
             </AccordionSummary>
-            <AccordionDetails className={linkedActivityProps.classes.formContainer}>
-              <FormContainer {...linkedActivityProps} />
+            <AccordionDetails className={props.classes.formContainer}>
+              <FormContainer {...{ ...props, activity: props.linkedActivity, isDisabled: true }} />
```
NickPhura, line 31: nit: can this be reduced to `{...props, activity: props.linkedActivity, isDisabled: true }` ?

I haven't quite nailed all of the different props syntaxes in my head, so maybe it will complain

app/src/features/home/map/MapPage.tsx
```diff
@@ -64,55 +64,27 @@ interface popOutComponentProps {
...
-            </Button>
-          </Grid>
+          {buttonLabels.map((item) => (
+            <Grid item key={item}>
+              <Button
+                disabled={item !== 'Close'}
+                color="primary"
+                variant="contained"
+                onClick={() => {
+                  props.buttonCloseCallback();
```
NickPhura, line 82: Will the onClick for these buttons always be the same? Or is this just the case currently because its a WIP?

app/src/components/activity/ActivityComponent.tsx
sdevalapurkar: yeah that was complaining when I tried earlier which is why I used the double spread :)

app/src/features/home/map/MapPage.tsx
sdevalapurkar: yeah this code was already present and I think it may be because it is WIP but not sure.... just thought of refactoring it to be cleaner. If the buttons need to be separated with different click functionality it can be updated later but if not then this will be cleaner was my thought

NickPhura (approved):

PR #438, opened 2021-01-17
Race condition

popkinj:
# Overview

- Refactoring SQL insert logic to occur in bulk. This fixes the Database Locking race condition witnessed before.
- Adding a test overlay which is connected to the offline functionality.
- Fixed a bug where the latitude and longitude were switched on data requests.

## Type of change

- [ ] Version change


micheal-w-wells (dismissed):

api/src/paths/context/databc/{wfs}.ts
```diff
@@ -91,6 +91,7 @@ function getDataBC(): RequestHandler {
     axios
       .get(url)
       .then((response) => {
+        //console.log('response from databc: ', response.data);
```
sdevalapurkar: Can we remove this line now?

api/src/paths/context/internal/{target}.ts
```diff
@@ -82,18 +82,30 @@ const getPlanningArea = async (lon: any, lat: any, res: Response, attr: string,
...
+  //   where
+  //     public.st_intersects(
+  //       public.geography(
+  //         public.st_setSrid(
+  //           public.st_point(${lon},${lat})
+  //         ,4326)
+  //       ),target.geog
+  //     )
+  // `;
+
```
sdevalapurkar: Same with this section of commented out code, can it be removed now?

api/src/utils/context-queries.ts
```diff
@@ -166,8 +174,8 @@ const saveInternal = (id: any, req: any) => {
  * @param req {object} The express request object
  */
 const saveElevation = (id: any, req: any) => {
-  const x = req.body.form_data.activity_data.latitude;
-  const y = req.body.form_data.activity_data.longitude;
+  const y = req.body.activity_data.latitude;
+  const x = req.body.activity_data.longitude;
```
sdevalapurkar: Instead of using x and y to name these can we just use lat and long as variable names? Would be maybe less ambiguous

sdevalapurkar (changes requested):

sdevalapurkar:
also I see the linting checks are failing, this needs to be fixed prior to merge :)

api/src/utils/context-queries.ts
```diff
@@ -98,8 +106,8 @@ const saveBCGW = (id: any, req: any) => {
  * @param req {object} The express request object
  */
 const saveInternal = (id: any, req: any) => {
-  const x = req.body.form_data.activity_data.latitude;
-  const y = req.body.form_data.activity_data.longitude;
+  const y = req.body.activity_data.latitude;
+  const x = req.body.activity_data.longitude;
```
NickPhura: I don't think `activity_data` is available from the body, its a property of `form_data`

`req.body.form_data.activity_data.latitude;`

api/src/paths/context/databc/{wfs}.ts
popkinj: Yeah totally.

api/src/paths/context/internal/{target}.ts
popkinj: Good eye @sdevalapurkar :)

api/src/utils/context-queries.ts
popkinj: Nice one @NickPhura ... Right you are.

popkinj: Hmmmm... I'm not seeing the `form_body` key anymore. This is the print out of request *req.body*.
```javascript
{
  activity_id: '9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d',
  activity_type: 'Observation',
  activity_subtype: 'Terrestrial Invasive Plant',
  date: '2019-04-12',
  deviceRequestUID: 'string',
  created_timestamp: '2018-11-13T20:20:39+00:00',
  geometry: [ { type: 'Feature', geometry: [Object], properties: {} } ],
  activity_data: { latitude: 49.29153, longitude: -124.95511, reported_area: 10 },
  activity_type_data: {
    negative_observation_ind: false,
    aquatic_observation_ind: false,
    primary_user_last_name: 'mike',
    secondary_user_first_name: 'mike',
    secondary_user_last_name: 'mike',
    species: 'banana',
    primary_file_id: 'test',
    secondary_file_id: 'test',
    location_comment: 'test',
    general_observation_comment: 'general comment',
    sample_label_number: 'string'
  },
  activity_subtype_data: {
    species: 'banana',
    distribution: 123,
    density: 123,
    soil_texture: 1,
    slope: 123,
    aspect: 123,
    flowering: true,
    specific_use: 123,
    proposed_action: 123,
    seed_stage: 123,
    plant_health: 123,
    plant_life_stage: 123,
    early_detection: 1,
    research: true,
    well_on_site_ind: true,
    special_care_ind: true,
    biological_care_ind: true,
    legacy_site_ind: true,
    range_unit: 'Canyon'
  }
}
```

api/src/utils/context-queries.ts
popkinj: Good point @sdevalapurkar .... However this is a common syntax for portraying coordinates.

sdevalapurkar: ok sounds good! was just wondering

api/src/utils/context-queries.ts
sdevalapurkar: how are you getting this to print? is it a request through postman?

popkinj: I'm printing to the console when an activity is submitted.

popkinj: It's the request body that the client is submitting to the API.

popkinj: Oh... I get what you're saying @sdevalapurkar 
Maybe I'm sending an outdated data package.

popkinj: Hmmmm.... I don't have postman running yet. Can someone give me an example data POST/PUT payload? That is current?

api/src/utils/context-queries.ts
NickPhura: Id have to side with shreyas on this one.  Mainly because the switch from 'latitude' to 'x' doesn't really benefit anyone.  If you were trying to track down where latitudes are used, or trying to decipher some code farther down the stack, seeing 'x' is not particularly informative.  Especially if you aren't a spatial developer who knows that 'x' means latitude.

api/src/utils/context-queries.ts
micheal-w-wells: I think I broke my postman setup.  A way to get up to date version is to fillout a plant observation here https://dev-invasivesbci.apps.silver.devops.gov.bc.ca/home/activities and then hit reviewed and sync on the 'my activities page' with the network tab open in chrome.  Look at request headers and scroll down.

Use istest1 so that you can sync, your idir has likely been wiped from dev via migrations.

NickPhura: Hmm, it must have been parsed out earlier in the code I guess?

That blob you pasted is what would normally be the content of the 'form_data' key.

api/src/utils/context-queries.ts
popkinj: Sure... I can make the switch.

api/src/utils/context-queries.ts
popkinj: Never mind... Just installed Postman.

NickPhura: This is the request body that is sent from the frontend:

```Typescript
{
    activity_id: activity.activityId,
    created_timestamp: activity.dateCreated,
    activity_type: activity.activityType,
    activity_subtype: activity.activitySubtype,
    geometry: activity.geometry,
    media:
        activity.photos &&
        activity.photos.map((photo) => {
        return { file_name: photo.filepath, encoded_file: photo.dataUrl };
        }),
    form_data: activity.formData
}
```

And then the form_data field should just have that blob you pasted before in it.

NickPhura:
> 
> 
> also I see the linting checks are failing, this needs to be fixed prior to merge :)

For the app, can just run `npm run lint:fix`.  It will lint all files, but if the linting is up to date (which I believe it is) it should only catch linting issues in newly changed files.  It will generally auto fix all of them, and list any it was unable to.

Though there might be a few lint items that we have chosen to ignore, for now? Correct me if I am wrong @sdevalapurkar 
Wait, starting to remember now, those ones just come up as warnings, rather than errors.

sdevalapurkar:
> > also I see the linting checks are failing, this needs to be fixed prior to merge :)
> 
> For the app, can just run `npm run lint:fix`. It will lint all files, but if the linting is up to date (which I believe it is) it should only catch linting issues in newly changed files. It will generally auto fix all of them, and list any it was unable to.
> 
> Though there might be a few lint items that we have chosen to ignore, for now? Correct me if I am wrong @sdevalapurkar
> Wait, starting to remember now, those ones just come up as warnings, rather than errors.

Yup all errors need to be fixed before merge. It is considered a blocker for PR. Warnings that can be addressed should be but not a blocker as some of lints suggestions for warnings cause app to break

api/src/utils/context-queries.ts
popkinj: Good thinking @micheal-w-wells ... That sounds like the best way to get the current schema.

popkinj:
Think I've fixed all the linting errors which I introduced.

sdevalapurkar:
There seem to still be failures here: https://github.com/bcgov/invasivesbc/pull/438/checks?check_run_id=1723476596#step:5:13

popkinj:
Yeah... I see that. I can't figure out what it's complaining about. Do you think the linter doesn't like the spaces?

sdevalapurkar:
> Yeah... I see that. I can't figure out what it's complaining about. Do you think the linter doesn't like the spaces?

Let me help you out

popkinj:
> > Yeah... I see that. I can't figure out what it's complaining about. Do you think the linter doesn't like the spaces?
> 
> Let me help you out

I just made the layer identical to the previous ones. Let's see how it does.

sdevalapurkar:
my most recent commit should make the action pass @popkinj

popkinj:
Hey, what linter extension do you guys use?

api/src/utils/context-queries.ts
popkinj: Thanks @NickPhura ... Yeah, I just verified the `form_data` is a thing.
I'm going to fix all the logic to match... Let's keep this branch going until I do.

sdevalapurkar:
should pass now @popkinj

api/test/observation-polygon.json
```diff
@@ -1,70 +1,81 @@
...
+      "aspect_code": "E",
+      "proposed_treatment_code": "MT",
+      "range_unit_number": "3",
+      "plant_life_stage_code": "MFD",
+      "plant_health_code": "GH",
+      "plant_seed_stage_code": "NO"
+    }
   }
-}
+}
```
sdevalapurkar:
```suggestion
}

```

sdevalapurkar (commented):
looks good to me other than nick's comments left to address regarding `form_data` and `x`, `y`, `lat`, `lng`

api/src/paths/context/internal/{target}.ts
```diff
@@ -89,11 +89,8 @@ const getPlanningArea = async (lon: any, lat: any, res: Response, attr: string,
       public.${table} "target"
     where
       public.st_intersects(
-        public.geography(
-          public.st_setSrid(
-            public.st_point(${lon},${lat})
-          ,4326)
-        ),target.geog
+        public.st_geographyFromText('POINT(${lon} ${lat})'),
```
NickPhura, line 92: Just double checking, there doesnt need to be a comma between lon and lat?

popkinj: Correct... haha. That one always get's me too @NickPhura . It's a text representation of spatial stuff called [Well Known Text](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry). Thanks for checking though.

popkinj:
Think we're all good here. Any last objections?

sdevalapurkar (commented):
I see 3 warnings about missing dependencies, does adding them break anything? or can they be added to comply with the lint checker? @popkinj

sdevalapurkar (approved):
Just a question about the missing dependencies in hooks, but otherwise LGTM!! 💯

PR #487, opened 2021-01-29
Bulk editing same Subtype

dogcomplex:
# Overview

- Adds limiting to multiselect for bulk editing to only allow one Subtype at a time

## Type of change

## How Has This Been Tested?

Tested locally

## Screenshots

2021-01-28 17_24_28-InvasivesBC - Search

## Checklist:


app/src/components/activities-list/SearchActivitiesList.tsx
```diff
@@ -100,7 +102,7 @@ const SearchActivitiesList: React.FC<ISearchActivitiesList> = (props) => {
           ActivitySubtype.Treatment_MechanicalPlant,
           ActivitySubtype.Treatment_BiologicalPlant
         ];
-        const bulkEditIsDisabled = !allowedBulkEditSubtypes.includes(activity.activitySubtype);
+        const bulkEditIsDisabled = editSubtype && activity.activitySubtype !== editSubtype || !allowedBulkEditSubtypes.includes(activity.activitySubtype);
```
sdevalapurkar, line 105: question:

do you think it would be better to enable the click but display an error when the subtypes arent the same?
cause if we keep the button disabled, it might not be super obvious to the user why the button is disabled/how to enable it. 

Also, for the references page, I have the create treatment button working the way specified above (so I think these two should be made the same for user experience, if we wanna go the disabled button route I can change that page functionality - or we can make this page work similar to that one)

dogcomplex:
Closing/deleting this PR as fresh changes have been made in here instead:

https://github.com/bcgov/invasivesbc/pull/509

PR #511, opened 2021-02-06
More form updates

sdevalapurkar:
# Overview

- Form updates to treatments
- Tooltip texts added

## Type of change

## How Has This Been Tested?

- Locally

### DIFF
--- a/api/src/openapi/api-doc.json
+++ b/api/src/openapi/api-doc.json
@@ -299,9 +299,6 @@
           "activity_data": {
             "$ref": "#/components/schemas/Activity_BulkEdit"
           },
-          "activity_type_data": {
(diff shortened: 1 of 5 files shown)


dogcomplex (approved):
Looks good.  Was confused by those Treatment_BulkEdit fields before anyway, so glad to see them streamlined

sdevalapurkar:
> Looks good. Was confused by those Treatment_BulkEdit fields before anyway, so glad to see them streamlined

The reason `Treatment_BulkEdit` was there was because treatment object had fields that needed to be enabled for bulk editing purposes before - within multiple different activity types. Now that field has been removed completely so it could be removed :)

Issue #514, opened 2021-02-08
Metabase Search Integration V2


ZoeSimon:
**Describe the Task**
@dogcomplex :

- Making the Metabase filtering a 2-way relationship by creating ability to push list of activities/POIs on Invasives side to Metabase
  - This will allow, in particular, geo-filtering of Metabase results - giving ability to pick specific points or filter to a geometric region.  Combined with Metabase filters, this gives full search power of both apps
- Retool the Metabase Question picker (on Plan My Trip page) to use a dropdown of questions pulled from API  (or, if failing that, tack on the name next to ID)

PR #542, opened 2021-02-22
Metabase Caching v2

dogcomplex:
# Overview

-  Trip page controls to select a Metabase query by name https://github.com/bcgov/invasivesbc/issues/514
-  Cached Activities page Create Query command to create a Metabase query using selected points, adding them to a set collection
-  Improvement to Cached Activities page to select points on click
-  Also highlights selected points on map
-  Improvement to Fetch button logic on PMT page to show Fetching status 

## Known Issues (TODO next):
- POIs are crashing this, likely due to sheer number.  Need to add limits to Metabase results, and/or reduce the data to just POI ids (via Metabase API guesswork)
- (missing feature) Cached Activities page is only currently working with metabase logic for Observations, but could work for Treatments and Monitoring as well as POIs.  Just need to change UI.

## Type of change

## How Has This Been Tested?

Local testing, running through the whole workflow with Metabase searching.  Tested with Activities - known issues still with POIs due to data overloads.

## Screenshots

Same as demo, though cleaned up Metabase queries list and added feature to collect queries in a designated Collection (set in ENV)

## Checklist:

- [X] have made corresponding changes to the documentation


dogcomplex:
Re: the code smells:
- GET/POST collection: same error appears with each of these endpoints, it's not recognizing the automatic use of these exports
- "expected an object to be thrown" - maybe, but every endpoint does this same thing
- TripDataControls megafunction - yeah definitely needs refactoring.  But I only added one more section to this messy thing, not a new crime
- createMetabaseQuery() - hmmm, I guess a bit complex but not quite sure how to simplify... adds overhead then...  any thoughts?  Care much?

micheal-w-wells:
Gave each smell a sniff and see where you are coming from.  Yes the trip data controls need a refactor - I might do that this week if I have time.  The api error you could probably address with one annotation or another.  Same deal with the error throws.

For now let's do this:  throw /* eslint-disable */  ontop of trip data controls and your metabase query file, and address the issues for the linter (not sonar cloud).  Those should be straight forward.  We'll refactor trip data controls soon enough.

dogcomplex:
Whoops, didn't notice the lint errors - those were a result of the first round of linting.  No need to ignore - will just fix.

Re: api error throws probably solvable with annotations - yeah probably something typescript can solve.  Iiiii - don't know off the top of my head though, and yeah I don't think it's worth the time to figure out right now, so if you have an idea go for it otherwise my vote's to wait - as this is just copy pasting the same structure every endpoint on the app uses which should have produced the same errors on every PR up til now.

Re: Trip Data Controls - yeah, definitely needs a big refactor, and not just style-wise - needs generalizing for POI/Activities.  But again, not urgent.  I'd say fix real bugs first there too.  So sounds good

micheal-w-wells (approved):

micheal-w-wells:
Delete the branch when you're ready champ

PR #544, opened 2021-02-22
Context Layers for the Map

popkinj:
# Overview

This PR includes the following new layers for the map:

- Ownership
- Wells
- Invasive Plant Management Areas
- Regional Invasive Species Organization Areas
- Gravel Pits
- Streams
- Wetlands
- Municipalities
- Regional Districts
- Natural Resource Districts
- Road Features Inventory
- Basemap Place Names

All have been configured for availability offline.


app/src/components/map/MapContainer.tsx
```diff
@@ -66,12 +77,81 @@ const MapContainer: React.FC<IMapContainerProps> = (props) => {
...
 
-  const getSteepSlopes = () => {
-    return L.tileLayer.offline('https://forest-bridges.s3.amazonaws.com/steep-areas/{z}/{x}/{y}.png', {
-      maxZoom: 24,
-      tms: true,
-      opacity: 0.5,
-      maxNativeZoom: 15
+
+  const getNRDistricts = () => {
+    return L.tileLayer.offline('http://localhost:8080/geoserver/gwc/service/tms/1.0.0/invasives:WHSE_ADMIN_BOUNDARIES.ADM_NR_DISTRICTS_SPG@EPSG:900913@png/{z}/{x}/{y}.png',{
```
micheal-w-wells: Don't we need to substitute the domain here based on some env var?

app/src/components/map/MapContainer.tsx
```diff
@@ -66,12 +77,81 @@ const MapContainer: React.FC<IMapContainerProps> = (props) => {
...
+      opacity: 0.8,
```
micheal-w-wells: Can you run vscode format / lint fix on MapContainer.tsx while you are in there?

app/src/components/map/MapContainer.css
```diff
@@ -2,20 +2,25 @@
   transform: scale(1.4) translate(0, 28px);
 }
 
+.leaflet-draw.leaflet-control .leaflet-draw-section:nth-child(2) {
+  transform: translate(0, -13px);
+}
+
 .leaflet-control-layers.leaflet-control {
-  transform: scale(0.955) translate(7px, 100px);
+  transform: translate(-7px, 20px);
```
micheal-w-wells, line 10: Did you try this mocking out a mobile layout?

micheal-w-wells (changes requested):

app/src/components/map/MapContainer.tsx
popkinj: Good catch @micheal-w-wells .... Yes, this should be an env variable.

app/src/components/map/MapContainer.tsx
popkinj: Sure.

app/src/components/map/MapContainer.css
popkinj: No I didn't. I'll see what it looks like on an iPad.

app/src/components/map/MapContainer.tsx
popkinj: Hmmm... Thinking maybe we should be proxying this through the API.

popkinj: That way we don't have to worry about the HTTPS-> HTTP issue. And we'll get the benefit of Keycloak.

micheal-w-wells: Until that is done will merging this break the app?

popkinj:
Hey @micheal-w-wells 
I've run the linter and made some updates to the URL logic to Geoserver. I'd like to eventually proxy Geoserver through the API. That would allow us to use environment variables and the like instead of forming the path in the front end.

app/src/components/map/MapContainer.tsx
popkinj: I couldn't get the vscode extension to work... So I installed the linter separately, ran it, and fixed all the complaints.

micheal-w-wells: either didn't get pushed up or your linter is different - seemed to have a bunch more issues than before.  I ran it for you

micheal-w-wells:
@popkinj your branch to delete when ready - remember to do a pull before you add more work to it if that was your plan.

popkinj:
Got it. Thanks @micheal-w-wells

Issue #540, opened 2021-02-19
Pre-user testing issues list


micheal-w-wells:
Crystal's gonna say it needs to happen, or will be annoyed with symptom of it not being done:
- _Broken into new ticket_ can we tell user there are too many in the area to fetch? -wk
- [x] add index to pouch query - mw 
- [x] batch insert into pouch - wk
- [x] point of interest bug - clicking POI on map - wk

_Broken into new ticket_ Would be nice:
- leaflet cluster markers
- finish POI photos
- clicking activities on map creates blank bottom menu - either fill that in with edit form / activity data, or disable it
- side-scroll bar on IAPP table styling (too dark)
- success/error message presentation - the stack is a bit clunky

_Broken into new ticket_ Maybe temporarily passable bugs
- Plan My Trip page controls sometimes unresponsive after Wipe Local Data pressed.  Need to switch tabs and wait
- (seems to appear on multiple pages, semi-randomly)  Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in ActivityListItem (at ReferenceActivitiesList.tsx:128)

PR #546, opened 2021-02-24
Pre flight bug fixes

dogcomplex:
# Overview

Fixes some bugs caused by too many metabase queries being thrown on fetches, though does not solve the underlying problem.  Sets API to only fetch on first load and when you click on the Metabase controls on Plan My Trip.

Underlying problem bug fix coming next... problem is the DatabaseChangesContext is updating once for every single change, even with a bulk edit, which is causing at minimum a big CPU/RAM lag spike on Fetch.  Any big import of thousands of points will cause a freeze.  Before this it was sending thousands of API calls to Metabase, crashing it.  Now it should just be a bad user experience as the app slows...

Note: the actual DB/API fetches don't seem to be the bottleneck - it's the changes feed which triggers thousands of times.

- https://github.com/bcgov/invasivesbc/issues/540

## Type of change

## How Has This Been Tested?

Locally with variety of fetches

## Checklist:


micheal-w-wells (approved):

micheal-w-wells:
your branch to delete when ready @dogcomplex

PR #598, opened 2021-03-03
Kml upload

popkinj:
# KML Upload Tool

Drag and drop KML files get added to the map. All attributes (meta data) area accessible in the map popup. If a KMZ file gets dropped... a message displays asking the user to unzip and provide the interior KML.

Special care was take to not clobber the existing KML upload logic in the *Plan my Trip* section.


micheal-w-wells (approved):

micheal-w-wells:
your branch to delete when ready @popkinj

PR #609, opened 2021-03-04
Updates to make apple happy

jleach:
- Updated `.gitignore` so that it ignores the build artifact directory `buildArchive`.
- Removed `buildArchive/options.plist` because it was stored in two places.
- Updated version and app name in `ios/App/App.xcodeproj/project.pbxproj`
- Added `ios/App/App/App.entitlements` with push notification support. Looks like the app registers for APN services so Apple was complaining this file was missing.
- Added a line to ` pack.sh` to copy `options.plist` in place.


micheal-w-wells (approved):
Thanks!!!

PR #648, opened 2021-03-18
Try fix for #637 issue

### LINKED ISSUE #637: Unable to view form for record created offline on iOS
**Describe the bug**
Works:  Making a new record and viewing form once while online, and then going offline and viewing it.
Doesn't:  Make a new record while offline, you'll get a spinner when you try to view the form.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

sdevalapurkar:
# Overview

- Try fix for issue where the form doesn't load when you are on ios (offline)
- The API spec is needed for form to load and the api spec does not get cached until you open a form in online mode - so now it has been changed so that the spec gets cached once user is online and authed so as long as they open the app once in online mode before going into the field, their spec will be cached in pouch and they should be able to create forms successfully in field (offline)
- #637 

## Type of change

## How Has This Been Tested?

- Yet to be tested by someone who has simulator or ios device (@micheal-w-wells possibly)

### DIFF
--- a/app/src/contexts/authStateContext.tsx
+++ b/app/src/contexts/authStateContext.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import { useInvasivesApi } from 'hooks/useInvasivesApi';
 
 export interface IAuthState {
   ready?: boolean;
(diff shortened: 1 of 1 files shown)


micheal-w-wells (approved):
Looks like it might solve it, I'll push out an update soon as I get a chance and myself or @mikeshasko will test.

sdevalapurkar:
> Looks like it might solve it, I'll push out an update soon as I get a chance and myself or @mikeshasko will test.

thanks! feel free to loop me into a call or something when you are starting the testing of it

PR #652, opened 2021-03-22
Fix map issues

sdevalapurkar:
# Overview

- Fix map issues with auto reset of extent and not allowing points to work properly, etc

## Type of change


app/src/components/map/MapContainer.tsx
```diff
@@ -308,12 +308,6 @@ const MapContainer: React.FC<IMapContainerProps> = (props) => {
     );
   };
 
-  const addZoomControls = () => {
```
micheal-w-wells, line 311: Not sure I'm understanding what you're trying to do here.  The user won't be able to change zoom with your changes unless I'm missing something.

sdevalapurkar: this function was unused and never called so I just removed it

micheal-w-wells: I see now i had a peek at your branch.  I'll let @popkinj take a peek at this one and and approve it - I think him and Mike were discussing some desired changes in behaviour.

Issue #651, opened 2021-03-22
Dropdowns reverting to blank state


ghost:
**Describe the bug**
With the Friday Mar19th release the dropdown formatting was changed to allow for autofill / select as one types. All this is terrific and appears to work as intended. 
However, when one leaves a given form, and returns, all the selections that were made on any (or all) of the dropdown revert to a blank "start" state. This is a critical error in the behavior.

**To Reproduce**
1. Go to any activity form and fill out one or more fields
2. Return to the my activities
3. Return to the form where the fields were entered

**Expected behavior**
Selections should be retained.

PR #653, opened 2021-03-22
Fix persistent changes bug and validation error problems patch

dogcomplex:
# Overview

- https://github.com/bcgov/invasivesbc/issues/651

Fixed the main bug (in the first 15 minutes, surprisingly), but then noticed some follow-up bugs which (inevitably) arose from it...

Long-story short, there is only one possible configuration here which gets roughly the behavior we want:  if a field is required, then we disable the ability to clear (make blank) the field once it has been selected (by pressing the little X button) because validation will not correctly read that and throw an error on blank inputs - allowing empty form fields.  That would be bad.

Meanwhile, since non-required fields might still want to be blank (e.g. when the person changes their mind and would rather say nothing than an answer they're not confident in) then we re-enable the ability to clear, and it doesn't matter that the validator doesn't catch it because it's not required.

Underlying problem still foreboding: is the validator actually noticing when the selected value is not one of the allowed options?  I'm guessing no right now, but not seeing how that could cause a breaking state... for now.  But this will rise again if/when there's some extra custom validation we need to trigger on an autocomplete field :)  My deepest hope is that when that time comes, another developer will stumble upon this old PR and say "hah, you were right Warren!  Too bad it's years later and you're not here to help attempt to solve yet another horrible layer of this problem!"

Lest we forget.
  

## Type of change

## How Has This Been Tested?

Locally

## Screenshots

## Checklist:


micheal-w-wells (approved):

micheal-w-wells:
Hilarious.  Let's try it and see how it works.

micheal-w-wells:
from the android failure:  


```
./node_modules/@ionic/core/dist/esm-es5/index-7a8b7a1c.js
Attempted import error: '__spreadArray' is not exported from 'tslib'

```

PR #743, opened 2021-05-19
Revert "Download"

micheal-w-wells:
Reverts bcgov/invasivesbc#736


micheal-w-wells:
@popkinj lmk if you need git help after this.  should be relatively straight forward to get your changes + fix in another branch.

PR #774, opened 2021-06-23
vector tile work

micheal-w-wells:
vector tile work - moving to separate files so easier to merge into react-leaflet


micheal-w-wells:
don't delete this branch just yet

PR #796, opened 2021-06-30
Geog to UTM Converter

micheal-w-wells:
* Create latLongToUTM function exported from WFSConsumer for converting Geog featurecollections to UTM.

- [ ] Needs code to get UTM zone from lat long
- [ ] Verify proj4 string
- [ ] maybe change it from featureCollection to just grabbing the lat long as input - a featurecollection could have many UTM zones which complicates things, and we have no need for the map to have GEOJSON in UTM.

### DIFF
--- a/app/src/components/map/WFSConsumer.tsx
+++ b/app/src/components/map/WFSConsumer.tsx
@@ -29,6 +29,31 @@ proj4.defs(
   'EPSG:3005',
   '+proj=aea +lat_1=50 +lat_2=58.5 +lat_0=45 +lon_0=-126 +x_0=1000000 +y_0=0 +ellps=GRS80 +datum=NAD83 +units=m +no_defs'
 );
+proj4.defs('UTM', '+proj=utm +zone=32 +datum=WGS84');
+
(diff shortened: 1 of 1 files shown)


app/src/components/map/WFSConsumer.tsx
```diff
@@ -29,6 +29,31 @@ proj4.defs(
   'EPSG:3005',
   '+proj=aea +lat_1=50 +lat_2=58.5 +lat_0=45 +lon_0=-126 +x_0=1000000 +y_0=0 +ellps=GRS80 +datum=NAD83 +units=m +no_defs'
 );
+proj4.defs('UTM', '+proj=utm +zone=32 +datum=WGS84');
```
micheal-w-wells, line 32: note to self remove this line

popkinj: Converting Longitude to a UTM zone:
```javascript
const utmZone = (Math.floor((longitude + 180) / 6) % 60) + 1;
```

micheal-w-wells:
Grisha used the formula in the comment to get it working in another branch - can close this now.

Issue #804, opened 2021-07-02
Add UTMs as well as lat/longs, or if only one can be there, should be UTMs. (also in 607).


hpertaia:
- [x] Add UTMs as well as lat/longs, or if only one can be there, should be UTMs.

Issue #801, opened 2021-07-01
Form Dialogs for temperature, wind, and herbicide fields


hpertaia:
- [x] If the user chooses the value that is out of appropriate range, display the dialog window asking if the user is sure, and if clicked to proceed, remove the error from the fields.

Issue #794, opened 2021-06-29
More form updates


micheal-w-wells:
- [x] #804
- [x] #801

_Originally posted by @micheal-w-wells in https://github.com/bcgov/invasivesbc/issues/695#issuecomment-870981307_

micheal-w-wells:
Pesticide User License & User record.
 
Pesticide User License field should be autofilled based on log-in (attached to user entering IDIR or BCeID) and then autofilled. Doesn't need to be on form, but needs to be recorded upon syncing/submitting record.

Issue #724, opened 2021-05-04
Prevent future dates from being chosen in all forms


CrystalChadburn:
- [x] user should not be able to pick a date past current date on all forms.  Default is today's date.

PR #805, opened 2021-07-02
issues: #804 (also in #794), #801 (also in #794), #724.

hpertaia:
# Overview

- [x] Add UTMs as well as lat/longs, or if only one can be there, should be UTMs. Added 3 new fields to the activity form: UTM Northing, UTM Easting, UTM Zone. Fields are populated automatically based on lat/long input.
- [x] Form Dialogs for temperature, wind, and herbicide fields. If the user chooses the value that is out of the appropriate range, display the dialog window asking if the user is sure, and if clicked to proceed, remove the error from the fields.
- [x] Prevent future dates from being chosen in all forms. All activity_date_time fields are now being validated when the user types in the value. If the current date/time is less than the value, the error is displayed.

## Type of change

- [ ] Version change

## How Has This Been Tested?

## Checklist:


micheal-w-wells (dismissed):
awesome!!  stoked you were able to get the projection transformation to work.

micheal-w-wells (approved):

PR #900, opened 2021-07-27
Myrecords review workflow and bug fixes merging

dogcomplex:
# Overview

- {List all the changes, if possible add the linked issue/ticket #}

## Type of change

- [ ] Version change

## How Has This Been Tested?

## Screenshots

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code

### DIFF
--- a/api/src/app.ts
+++ b/api/src/app.ts
@@ -20,7 +20,10 @@ app.use(function (req: any, res: any, next: any) {
   defaultLog.info(`${req.method} ${req.url}`);
 
   res.setHeader('Access-Control-Allow-Credentials', true);
-  res.setHeader('Access-Control-Allow-Headers', 'X-Requested-With, Content-Type, Authorization, responseType');
+  res.setHeader(
+    'Access-Control-Allow-Headers',
+    'X-Requested-With, Content-Type, Authorization, responseType,Access-Control-Allow-Origin'
+  );
   res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS, PUT, PATCH, DELETE, HEAD');
   res.setHeader('Access-Control-Allow-Origin', '*');
   res.setHeader('Cache-Control', 'max-age=4');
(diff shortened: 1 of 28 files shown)


micheal-w-wells:
Hey @dogcomplex, hopefully these notes help you sort this out:

For better or worse - getCachedAPI spec was updated in useInvasives API & not useDataAccess in the previous PR that touched this stuff  This would have been ok based on how you spread ...api in useDataAccess exports, but then you also called useDataAccess from invasivesAPI.getCachedAPISpec which must be causing a loop unless I misread.  If you just remove the reference to useDataAccess in useInvasivesAPI that is fine for this PR, can clean it up in another one.  No big deal.

To follow up on your comment yesterday the 'hostname not found' error was fixed in dev around lunchtime yesterday and should be fine here, we sent a note on Discord yesterday when we were working on it, if you are still seeing it after addressing the above issue I'd do a clean checkout of dev and see if you're still impacted.  I tested this locally and it's also working in Openshift.

It also looks like you clobbered the biocontrol form.  I'd really recommend you use something like git add -p to eyeball what you stage so you can intervene when necessary.  If it was during a rebase you likely would have had to accept both incoming and current as you both added tables.  You'll want to test that this form works & can save & submit via local api once you fix that as you have both client and api side changes in the PR.  This also could MIGHT happened if you did a git pull & didn't have ff only set in your config, but I suspect it was from one of the other commands.

Lastly & this is definitely a request for me:  though tempting I'd try not to run the linter for the whole repo - now you've got a bunch of other files like the map that should myself or Brian merge anything in is only going to cause you more grief on rebasing from dev.  I'll follow up with Brian's linter setup & hold out until 10am before i merge anything else, but in general 'lint only files changed' should get cleaned up by the last author in another pr, this also makes reviewing stuff WAY faster.  Github doesn't have a 'lint only' indicator for files which would negate the issue so we have to make a point of not adding a bunch of irrelevant changes.  I think this will end up saving you some pain with bigger PRs when you try to update your branch.

Especially in the stages prior to a decent UI test harness / PR smoke test, we've got to be pretty surgical, yes I'm being picky but it's just to make sure we as a team don't repeat previous mistakes.

PR #1106, opened 2021-09-22
Sync up test env with dev

micheal-w-wells:
Get code up to date in test.


micheal-w-wells:
closed - rolling with 1107

PR #1164, opened 2021-09-29
Spatial seeds

popkinj:
# Spatial Seeds
The following tables contain enough data they require external storage. The KnexJS migration tool seeds these tables by requesting and executing Postgres backup scripts. When tables are too large for NodeJS to store in memory, they are split into separate statements and executed individually.

Here are the new tables:
- regional_invasive_species_organization_areas
- invasive_plant_management_areas
- aggregate
- jurisdiction


popkinj:
Hey @micheal-w-wells 
I had lots of trouble with merge conflicts doing the rebase. For some reason there are redundant log entries. Not sure how that happened. But I believe everything is there. We'll definitely need to do a Squash Merge.

micheal-w-wells:
Hi @popkinj .  If you have a conflict on a  rebase often it will ask for a conflict commit but will autofill with the last commit message.  These happen most often with a really stale branch.

What you can do after the rebase is ```git reset --soft dev && git commit -m 'new squashed commit message'```.  This will turn it all into one commit, but you need to have both the rebase done and dev up to date locally when you run this.

The other trick to dodge the conflicts altogether (especially when your changes are seemingly so unrelated is to make a new up to date branch, but copy your changes over with ```git checkout oldbranch -- filepath/filename```.

Regarding you still getting a lot of conflicts I would definitely look into your local gitconfig and make sure you have ff only on pull set.  I have a feeling that one might be biting you.

micheal-w-wells (approved):

PR #1179, opened 2021-10-05
Now using the Ministry S3 for all largee seed files.

popkinj:
# Overview

- Seeding large datasets now reference a Ministry owned S3 Bucket
- Documentation for managing S3 objects


micheal-w-wells (approved):
sweet!!

PR #1247, opened 2021-10-27
Pmp

popkinj:
# Jurisdiction Migration Fix
- Fixes the database migration errors.
- Data within Jurisdiction migration has been moved to a seed file.
- Seed files in S3 have been modified (for Jurisdiction and Aggregate tables) to exclude schema changes.
- Tests on migrations passed.
- Tests on seeds will follow in a lower priority PR.


micheal-w-wells (changes requested):
can you throw 'drop table if exists' statements in the down migrations?  that way people working on their own migrations won't get stuck if they end up rolling it back.  otherwise LGTM

popkinj:
Oh you mean if you don't put the "if exists" part.... it complains?

micheal-w-wells:
> Oh you mean if you don't put the "if exists" part.... it complains?

From what I saw there are some migrations that just have empty 'down' functions, that means when someone does knex migrate:rollback nothing will happen, and then when you re-run migrate it will fail.

Putting 'drop table x if exists' in down migrations is just standard practice and pretty bomb proof.

popkinj:
Right... got that part.
I did have drop statements in the two migrations that are ready to go. Aggregates and Jurisdictions. The other two new files are just blank until I fill them.

popkinj:
Guess I should have mentioned that in the description.

database/src/migrations/20211027114020_invasive_plant_management_areas.ts
```diff
@@ -0,0 +1,10 @@
+import * as Knex from "knex";
+
+
+export async function up(knex: Knex): Promise<void> {
```
micheal-w-wells: sorry to nitpick but these unfinished migrations won't rerun in test or prod when you actually write them, probably want to exclude them from PR.

popkinj:
Oh good point. I'll get ride of them for now.

popkinj:
Also... There is the old migrations (that was trying to load data). That I left in place... gutted. Should I remove this as well?

micheal-w-wells (approved):

micheal-w-wells:
If you aren't going to edit them again it doesn't really matter if they are gutted.  If you do remove them everyone will have a manual step to make knex happy so just leave em in for now.

popkinj:
That's kinda what I was worried about... I'll leave in place then.

PR #1328, opened 2021-11-29
Mech query

popkinj:
# Mechanical Treatment Query
- New table generation SQL for splitting species based on Mechanical Treatment Query
- Stats SQL statement for creating Metabase "Question"

### DIFF
--- a/api/test/sql/spatial-link-observations.sql
+++ b/api/test/sql/spatial-link-observations.sql
@@ -1,0 +1,209 @@
+/* spatial-link.sql
+  Logic for spatially linking records of the same species
+  within a certain distance from eachother.
+ */
+
+/*
+  Deletes
+  1. Create an exploaded layer of positives
+  2. Create an exploaded layer of negatives
+*/
+-- Positive Layer
+drop table if exists test_spatial_expload_positive;
+create table test_spatial_expload_positive as
+-- create temporary table test_spatial_expload_positive as
+  select 
+    activity_type,
+    activity_subtype,
+    created_timestamp,
+    activity_incoming_data_id,
+    jsonb_array_elements(to_jsonb(species_positive)) "species", -- Explode species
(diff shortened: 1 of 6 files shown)


api/test/sql/spatial-link-observations.sql
```diff
@@ -10,7 +10,8 @@
 */
 -- Positive Layer
 drop table if exists test_spatial_expload_positive;
-create temporary table test_spatial_expload_positive as
+create table test_spatial_expload_positive as
```
micheal-w-wells, line 13: I get this isn't a migration (so it won't block pipeline), but just a heads up on the typo:  I think you meant ```create view``` and also I think the query would need to be wrapped in braces like ```create view view_name as ( ...query...);```

popkinj: Thanks @micheal-w-wells 

It has to be a table to accept the spatial index unfortunately. There's going to be two tables we'll need to create that all report views/queries will run off.
- treatments_by_species
- observations_by_species

When these tables get built they require indexes on the temporary intermediate tables. It makes for snappy run times... But let me know if there's another way to create spatial indexes. ☺

micheal-w-wells: Hey no worries was mostly just pointing out the syntax.  I haven't had any issues with spatial indexes other than the need to wrap the spatial functions with immutable ones for index to be used in functions, and that jsonb/spatial combo indexes seem to be too big to be used much of the time, so I get the need for another level/set of tables from that perspective, otherwise they do seem to work in views (we are using them that way now).

popkinj: Hey @micheal-w-wells .
Do you want me to make any changes here?

popkinj: Forgot about the logs... They're gone now.

micheal-w-wells (approved):

PR #1351, opened 2021-12-09
Chemical Treatment Query

popkinj:
# Chemical Query is working
Currently the SQL reports the following
- Species
- Treatment type
- PMP boundary

The treatment type can be easily altered to other forms like: pesticide, method etc...


app/src/hooks/useInvasivesApi.ts
```diff
@@ -390,6 +390,7 @@ export const useInvasivesApi = () => {
   const updateActivity = async (activity: ICreateOrUpdateActivity): Promise<any> => {
     // Not sure who is using this... But its smelling
     // const oldActivity = await getActivityById(activity.activity_id);
+    console.dir('updating',activity)
```
micheal-w-wells: hey champ if you can remove this line we'll merge it in

micheal-w-wells (changes requested):

micheal-w-wells (approved):

PR #1469, opened 2022-01-10
Roles Backend

sam-warren:
# Overview

- Updated test users seed to add some default roles
- Added user access, user roles tables
- Add queries for user access table
- Add front end api functions to query user access 

## Type of change

- [ ] Version change

## How Has This Been Tested?
Manual testing

## Checklist:


database/src/migrations/20211214033115_access_request.ts
```diff
@@ -1,11 +1,10 @@
 import * as Knex from 'knex';
 
 export async function up(knex: Knex): Promise<void> {
-    await knex.raw(`
-    set schema 'invasivesbc';
+  await knex.raw(`
+    set schema 'public';
```
micheal-w-wells: When you set a search path, you're telling psql "if I give a relation without a schema prefix, FIND it via these schemas in this order", when you give it a schema, i BELIEVE it sets the schema as where you want to CREATE a table if you don't give a schema prefix, and also doubles as a search path if you didn't give one.  Long story short I think you want to set the search path to invasivesbc or just prefix your table/relation names with invasivesbc.

database/src/migrations/20220106120516_user_roles.ts
```diff
@@ -0,0 +1,47 @@
+import * as Knex from 'knex';
+const DB_SCHEMA = 'invasivesbc';
```
micheal-w-wells, line 2: this one is probably fine

database/src/migrations/20220113030811_access_request_patch.ts
```diff
@@ -0,0 +1,51 @@
+import * as Knex from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set schema 'public';
```
micheal-w-wells: this one will put access_request in public

docker-compose.local.yml
```diff
@@ -40,6 +40,7 @@ services:
       - local-network
     links:
       - db
+    restart: on-failure
```
micheal-w-wells, line 43: nice!

micheal-w-wells (changes requested):
Awesome work champ.  I think you just need to take a look at my notes for migrations and otherwise (assuming it was tested on mobile too) it should be good to go.

database/src/migrations/20211214033115_access_request.ts
sam-warren: fixed!

PR #1495, opened 2022-01-21
WFS Crash Fix

popkinj:
# WFS Crash Fix
The map was crashing whenever DataBC sent an empty feature array. I added a couple try/catch statements that fix this.
I should also add that there is a Pesticide query mostly baked. It has no effect on either backend or frontend yet.


api/test/sql/jamies_test_data.sql
```diff
@@ -0,0 +1,13 @@
+drop table if exists jamies_test_data;
```
micheal-w-wells: i thnk we can leave this one out

app/src/hooks/useInvasivesApi.ts
```diff
@@ -452,6 +452,7 @@ export const useInvasivesApi = () => {
   const updateActivity = async (activity: ICreateOrUpdateActivity): Promise<any> => {
     // Not sure who is using this... But its smelling
     // const oldActivity = await getActivityById(activity.activity_id);
+    console.log(activity);
```
micheal-w-wells: this can get pretty noisy

micheal-w-wells (changes requested):

micheal-w-wells (approved):

PR #1809, opened 2022-04-19
Cache Dropdowns

sam-warren:
# Overview

- Moved dropdown caching to useDataAccess.tsx, cache dropdowns on login since they are user-specific
- Fixed upsert where ID was not a primary key, drop code_tables table on login since different users see different tables.
- Tested blue dot tracking on mobile - working
- Tested browser cache clear - non-issue
- Updated makefile to make building and running ios simpler (`make ios`)

## Type of change

- [ ] Version change

## How Has This Been Tested?

Testing on mobile

## Checklist:

### DIFF
--- a/app/src/contexts/DatabaseContext.tsx
+++ b/app/src/contexts/DatabaseContext.tsx
@@ -143,7 +143,6 @@ export const DatabaseContextProvider = (props) => {
     for (const value of enumKeys(DocType)) {
       switch (value) {
         case 'CODE_TABLE':
-          setupSQL += `DROP TABLE IF EXISTS ${DocType[value]};`; // To wipe old table and get latest codes
           setupSQL += `create table if not exists  
             ${DocType[value]} 
              (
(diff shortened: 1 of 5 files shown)


app/src/contexts/DatabaseContext.tsx
```diff
@@ -143,10 +143,11 @@ export const DatabaseContextProvider = (props) => {
     for (const value of enumKeys(DocType)) {
       switch (value) {
         case 'CODE_TABLE':
+          setupSQL += `DROP TABLE IF EXISTS ${DocType[value]};`; // To wipe old table and get latest codes
```
micheal-w-wells: I see what you're going for but I think what might happen is the code tables getting wiped on startup while out of network.

micheal-w-wells (commented):
Looks good but that one gotcha I think we want to address before we merge.

app/src/contexts/DatabaseContext.tsx
sam-warren: Removed

PR #1940, opened 2022-06-02
Updated API Versions

rstens:
# Overview

- Updates to the DCs for API and APP - The horizontal scaler API Versions

## Type of change

You might have noticed errors for app and api HPA reported: **no matches for kind "DeploymentConfig" in group ""**

This has to do with OpenShift been upgraded and therefore some of the API version that we are using in the deploy are now deprecated. This change updates 2 api version in the HPA definition so it can properly operate again.


micheal-w-wells (approved):
THANK YOU

Issue #2304, opened 2022-11-18
Ensure IAPP and InvasivesBC records are showing on the public facing map again


CrystalChadburn:
Bring back the iapp and invasivesBC records on the public facing map before you log in.

Bring back the iapp and invasivesBC records on the public facing map before you log in.

Techy todos:
-[ ] 

How activities page/activity page + map works

1. user logs in / stays logged out
2. auth request complete
3. app fetches geojson for both activities and points of interest / iapp stuff --- works up to here logged out
4. map init action
      grabs all the filters for each record set from local storage
      puts filters into state (usersettings.recordsets)
      loop over each recordset, and fetch the IDS ONLY for that recordset
        activities endpoint (id only flag)
        points of interest endpoint (site_id only flag)

recordsetlayersrenderers - just filter the 2 big feauturecollections by recordset IDLists - no new server fetches

5. user clicks on recordset - iapp tablerows get || activity tablerows get - these will get full rows for display
     these actions call the same endpoint, 
    but without the id only flag, and use the same filters, but pass paging variables, ans sorting keys

what needs to change for public:

- [ ] check role / logged in before calling admin shape endpoint
- [ ] hide the boundary thing for now - nice to have when other stuff done - put it back, but dont upload or pull from server 
- [ ] activities and points of interest endpoints:
      need to work while logged out BUT:
        - [ ] can't return person name && can't filter by any name
        - [ ] dont worry about hiding columns - just dont pull them from server
- [ ] activiy page - can disable the activity page saga while logged out  
- [ ] other misc ui stuff:
    - [ ] hide the new record button
    - [ ] double check tabs make sense for public

micheal-w-wells:
This issue is more of an 'epic' - The team shifted to working on getting the new layer format to work before doing this to avoid extra rework.  Doing smaller increments/tickets to get there, will keep this ticket around for info but not drag through sprint point tracking as its too broad.

PR #2405, opened 2023-01-20
public map authentication logic

micheal-w-wells:
# Overview

- #2304 portion for POI public API

## Type of change

- [x] Added isAuth authContext property to indicate if a user is authenticated or not.
- [x] Split original request URL by the query symbol(?) and match URL path with public allowed paths.  If successful then set isAuth to FALSE.
- [x] If authorization header is well formed and matches the token then we set isAuth to TRUE.
- [x] Get Points of Interest now publicly returns all results. This can be adjusted if restricted fields are defined.

## How Has This Been Tested?

Manual API testing.  Did not have data for points of interest in database to test but returned 200 on results.

## Screenshots

image
image
image
image

## Checklist:

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -1405,6 +1405,25 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/array.prototype.reduce": {
+      "version": "1.0.5",
+      "resolved": "https://registry.npmjs.org/array.prototype.reduce/-/array.prototype.reduce-1.0.5.tgz",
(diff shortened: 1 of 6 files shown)


chrisrafusetech:
@micheal-w-wells  Could we just set security to [] and remove the public url check for POI API endpoint or keep as is?
```
GET.apiDoc = {
  description: 'Fetches all points of interest based on search criteria.',
  tags: ['point-of-interest'],
  security: SECURITY_ON
    ? [
        {
          Bearer: ALL_ROLES
        }
      ]
    : [],
```

micheal-w-wells:
While it's looking like we did a better scrub than thought of data to filter out when we imported everything, we'll still take these changes because we can use the user info for logged in users to aid in caching result sets serverside, and if we get asked to only publish records after a certain period of time for public then we're well set up.

Will need this branch updated to work with Rob's recent changes.

micheal-w-wells:
this worked merged into dev via public_map_test branch

Issue #3143, opened 2024-02-13
MapLibre - Recordset layers v1


micheal-w-wells:
- [x] Layers toggleable 
- [x] Layers update with tables
- [x] Layers change colour
- [x] Layers have labels
- [x] IAPP and IBC

PR #3152, opened 2024-02-20
Maplibre geometry edit for activities - part 1: loading intitial geo as editable

micheal-w-wells:


micheal-w-wells:
Most of #3143 , needs pin drop, buffer line, and extraneous buttons removed.

PR #3394, opened 2024-07-30
[INV-1289] "Track Me" Geometry drawing

### LINKED ISSUE #1289: GPS tracker fix (to draw the polygon for your record)
- [ ] ensure that the GPS on the ipad is communicating accurately with InvasivesBC
- [ ] enable the "track polygon" using the GPS on the forms to work (ask @micheal-w-wells  how to do this)
ONLY IF DONE IN ONE DAY.

LocalNewsTV:
# Changes

## Notification System
- Creation of New Alert System Using interface `AlertMessage` with MUI's `Alert`
- Notification System overlays application in top right corner, replacing the existing `SnackBar` Alert system
- Alert system is customizable for Icons, Severity, Titles, Text, and Timeout function, using Enums for enforcement
- Measures taken to allow alerts to self-delete and avoid duplicate entries
- Users can manually clear a notification, or Clear ALL notifications

Example Icons:
image

## Track Me Draw Geo
- Added Mapping Notifications to help guide users on possible errors in their GPS drawn icons
- Enforced a minimum distance between points of 10 meters to minimize spamming points on shape.
  - This size is about 2x the size of [average GPS accuracy](https://www.gps.gov/systems/gps/performance/accuracy/#how-accurate)
  - May have to increase value after testing on a few devices to gauge effectiveness
- Closes #1289

Running in Mobile Mode, iPad Air Dimensions, Edge Browser
image

## Misc
- Shifted most Geo Shapes I came across into a GeoShapes Enum for typo minimizing cross platform.
 
### Known Error(s)
- If user walks a perfectly straight line from `A`->`B` and closes the track tool, it will abort. This also occurs when using the Polygon draw tool, and may not actually be a bug, but just an inconvenience.
  - The GPS would have to record a user going perfectly straight for at least 20 meters with no deviation on one Axis
     - This is much easier to replicate using Browser sensors, of course.
  - Unfortunately without doing collinear equations, I cannot find a way catch this specific use case


app/src/UI/AlertsContainer/AlertsContainer.tsx
```diff
@@ -0,0 +1,97 @@
...
+import './AlertsContainer.css';
+import AlertMessage from 'interfaces/AlertMessage';
+import { CLEAR_ALERT, CLEAR_ALERTS } from 'state/actions';
+import { AlertSubjects } from 'constants/alertEnums';
+import { useEffect, useRef } from 'react';
+
+const AlertsContainer = () => {
+  const dispatch = useDispatch();
+  const timeoutsRef = useRef<ReturnType<typeof setInterval>>({} as ReturnType<typeof setInterval>);
+  const alerts = useSelector((state: any) => state.AlertsAndPrompts.alerts || []);
```
plasticviking: We have a type `useSelector` available in `utils/use_selector.ts`. Preferable to `any`.

app/src/UI/Map2/Controls/FindMe.tsx
```diff
@@ -1,9 +1,15 @@
...
+  MAP_TOGGLE_TRACK_ME_DRAW_GEO_START,
+  MAP_TOGGLE_TRACK_ME_DRAW_GEO_STOP
+} from 'state/actions';
 import 'UI/Global.css';
 import MyLocationIcon from '@mui/icons-material/MyLocation';
+import PolylineIcon from '@mui/icons-material/Polyline';
+import DirectionsWalkIcon from '@mui/icons-material/DirectionsWalk';
 
 export const FindMeToggle = (props) => {
   const positionTracking = useSelector((state: any) => state.Map?.positionTracking);
```
plasticviking: Here too

app/src/UI/Map2/Controls/FindMe.tsx
```diff
@@ -42,7 +47,56 @@ export const FindMeToggle = (props) => {
     </div>
   );
 };
+/**
+ * TrackMeButton
+ * @description Component to handle the functionality of the find me button
+ * @returns {void}
+ */
+export const TrackMeToggle = (props) => {
+  const positionTracking = useSelector((state: any) => state.Map?.positionTracking);
```
plasticviking: here

app/src/UI/Overlay/Records/Activity/Form.tsx
```diff
@@ -35,6 +38,7 @@ export const ActivityForm = (props) => {
   } = useSelector((state: any) => state.ActivityPage?.activity);
 
   const invasive_plant = useSelector((state: any) => state.ActivityPage?.activity?.invasive_plant);
+  const drawGeometryTracking = useSelector((state: any) => state.Map?.track_me_draw_geo);
```
plasticviking, line 41: ...and here. Though here I see the untyped useSelector is already in use by existing code. It may or may not be straightforward to change. Also Map can't ever be null, so that ?. doesn't do anything.

app/src/utils/geomWithinBC.ts
```diff
@@ -0,0 +1,26 @@
+import { BC_AREA } from 'sharedAPI';
```
plasticviking: I'd been specifically using the dynamic import to avoid pulling a large JSON blob into our bundle. It makes a noticeable difference in load time and is data, not code. I think it makes sense to continue to use dynamic import unless there's a compelling reason we shouldn't.

LocalNewsTV: @plasticviking  I had read the original function I lifted out as a "First check, then dynamic" so when I migrated it I attempted to preserve it with the context of being solo. I will update it to only the dynamic import :)

plasticviking (approved):
lgtm

PR #3397, opened 2024-08-01
[INV-3391] Simplify API return for Activity Autofill

### LINKED ISSUE #3391: Modify endpoint for user autofill
Modify existing endpoint that is used for the suggestedPersons autofill

LocalNewsTV:
# Overview

## Modify the `get` method for `/application-user`:
- Block non-app users (users with 0 roles)
- Create a sanitized set of data for non-admin users.
  - Currently only returns values needed for PAC Autofill
    - Can be easily adjusted later as needs change


plasticviking (approved):
Looks good to me

PR #3436, opened 2024-08-22
[INV-3389] Hide Draw tools when not needed

### LINKED ISSUE #3389: Draw tools should be hidden when just viewing the map
They should appear when drawing a drawn filter or creating a new record

LocalNewsTV:
# Overview

- Condensed logic in RefreshDrawControls to reduce complexity. 
- Added `Alert` when turning on the `What's Here?` Feature
    - The Feature already doesn't use the draw tools, users can just draw directly on the app, so notify them of this.
 - Removed some commented out code that looks like it was reintegrated elsewhere in the app

image

**When can you expect to see draw tools**
- When an Activity is open
- When you are drawing a custom layer
- Closes #3389

> [!WARNING]
> Bug still exists that causes Draw tools to sometimes duplicate (Part of me feels this is just caused in dev environments from rerenders of the map)

### DIFF
--- a/app/src/UI/Map2/Helpers.tsx
+++ b/app/src/UI/Map2/Helpers.tsx
@@ -159,7 +159,7 @@ export const mapInit = (
         public_layer: {
           type: 'vector',
           url: `pmtiles://${PMTILES_URL}`,
-          attribution: 'Powered by ESRI',
+          attribution: 'Powered by ESRI'
         }
       },
       layers: [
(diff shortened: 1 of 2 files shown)


app/src/UI/Map2/Helpers.tsx
```diff
@@ -569,39 +569,34 @@ export const rebuildLayersOnTableHashUpdate = (storeLayers, map, mode, API_BASE)
...
-      map.removeSource(source)
-    } catch(e) {
-      console.log('error removing source', e)
+      map.removeSource(source);
+    } catch (e) {
+      console.log('error removing source', e);
     }
-  })
-
+  });
```
LocalNewsTV, line 600: Just Prettier changes

plasticviking (approved):

PR #3446, opened 2024-08-27
[INV-3437] What's Here updates

### LINKED ISSUE #3437: Fix / Enhance "What's Here" functionality
# Describe the task

Did not properly understand how to use the Whats Here tool
~~The What's Here tool seems to be not working correctly~~, Users can select an area and receive a payload from the API, but the information isn't getting displayed.

UI Enhancement work can also be done to the Whats Here panel 

## Value Add
- Functionality works as expected
- More professional Look and Feel

## Acceptance Criteria

- [ ] Functionality Works as Expected
- [ ] UI Cleaned up for better look and feel

LocalNewsTV:
> [!TIP]
> The diffs look massive, but a large part of it was moving the `Helpers` components into new sub component files.

# Overview

- Separated Whats here `Helpers` into sub-components, better aligning with Vite's performance of 1 export per file
    - Also reduces complexity in parsing through file when changes are needed
- Display more detailed message when 0 records are displayed
    - Provides a direct link back to the `/Records` page so users can ease navigation
- Datagrid now pads out columns to better display data, and scales across devices
- Converted renderCell functions to implicit return over explicit.
- Removed a lot of conditional rendering around rendering disabled buttons, instead applied logic to disabled attribute of the buttons
- Change Theming of 'Whats Here' for greater consistency across application
    - Lots of BC Blue was being used, blending elements together in the view
- Closes #3437 

## New

image
Message displayed on empty rows

image
Table with rows

## Old

image
Massive amounts of whitespace left, and results were capped at 5 rows per page

image

When scroll is applied


plasticviking (approved):
looks nice

PR #3447, opened 2024-08-27
[INV-3443] Remove BR Tags

### LINKED ISSUE #3443: :zap: Remove `br` tags from app tables
# Describe the task

`br` tags constantly are flagged and errored for being improperly used inside of `td` tags. Hardcoding these html elements is bad for styling, and in the flagged use cases, aren't really doing anything

## Value Add
- Simplifies the codebase
- reduces errors being thrown

## Acceptance Criteria

- [ ] Errors are no longer being thrown around invalid DOM nesting. 
- [ ] Everything still looks the same or better on PC/iOS

## Additional Context

Image

LocalNewsTV:
# Overview

- Convert Divs being created using `map` to unordered lists where appropriate
- Remove all BR Tags in HTML as a small step towards better [Accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility)
    - Elements using tags had styling adjusted to maintain the same look as before.

- Closes #3443

### DIFF
--- a/app/src/UI/Map2/LayerPicker.css
+++ b/app/src/UI/Map2/LayerPicker.css
@@ -6,3 +6,9 @@
   text-wrap: wrap;
   box-sizing: border-box;
 }
+
+.layerPickerList {
+  list-style-type: none;
+  padding: 0;
+  margin: 0;
+}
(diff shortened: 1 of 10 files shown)


plasticviking (approved):
we are very inconsistent about using MUI typography vs html tags. ideally we should standardize on one or the other eh? generally I think for our application it makes more sense to prefer HTML tags since we're unlikely to ever go to a non-HTML rendering solution

PR #3502, opened 2024-09-23
[INV-3500] Tech Debt**  Change console.logs to warn/info/error, delete unneeded ones

LocalNewsTV:
# Overview

This PR includes the following proposed change(s):
- Refactor console.logs in `catch` statements to be `console.error`
- Wrap console.logs in debug statements to be consistently { wrapped }
- Refactor some console logs to be `.info` and `.warn` statements where appropriate
- Delete unneeded `.log` and `.dir` statements to reduce console bloat 

- Closes #3500

PR #3512, opened 2024-09-23
[INV-3505] Enhance Recordset UI

### LINKED ISSUE #3505: Enhancement - Recordset UI
# Describe the task

Update the UI for Recordsets so the background is the chosen recordset colour instead of border, remove the border effect, decouple the CSS from the `.tsx` file where possible

## Value Add
Better visual appeal, reduces border clutter. Works for users both on Mobile and PC

## Acceptance Criteria

- [ ] Background displays recordset color with lower opacity

LocalNewsTV:
> [!CAUTION]
> ~~This PR Is branched off of #3502~~
> This PR is up to date

# Overview

- Convert useSelectors to `Typed` version
- Reduce Ternary's in Jsx that resulted in React Fragments to be simple `(!)condition && Component` statements
- Remove Route wrapping component (tech debt item?)
- Increase `Touch Target` size for Recordsets `30px` -> `50px`
- Remove `type: any` from functions, convert to proper types.
- Swap most buttons for `IconButton` 
- Move most inline CSS to CSS File (Duplicated in iAPP since there is some crossover in areas of app)
    - Kept backgroundColor CSS properties, due to programmatic effect
    - removed border colouring for recordsets in favour of setting the background to lower saturation version of colors
        - Mobile users can't really benefit from `Hover` pseudo classes like desktop users can
- Closes #3505 

## Screenshots

Old Version

Border hover effect creates thick border
image

image

New Version 

All colors displayed

image

Hover increases Saturation

image

image

### DIFF
--- a/app/src/UI/Overlay/Records/Records.tsx
+++ b/app/src/UI/Overlay/Records/Records.tsx
@@ -49,8 +49,10 @@ export const Records = () => {
   const MapMode = useSelector((state) => state.Map?.MapMode);
   const recordSets = useSelector((state) => state.UserSettings?.recordSets);
 
-  const [loadMap, setLoadMap] = useState({});
+  const [highlightedSet, setHighlightedSet] = useState<string | null>();
   const [isActivitiesGeoJSONLoaded, setActivitiesGeoJSONLoaded] = useState(false);
+  const [isEditingName, setIsEditingName] = useState(false);
+  const [loadMap, setLoadMap] = useState({});
 
   const history = useHistory();
   const dispatch = useDispatch();
@@ -206,8 +208,8 @@ export const Records = () => {
     }, 3000);
   };
 
-  const [highlightedSet, setHighlightedSet] = useState<string | null>();
-  const [isEditingName, setIsEditingName] = useState(false);
+  const highlightSet = (set: string) => setHighlightedSet(set);
+  const unHighlightSet = () => setHighlightedSet(null);
 
   if (!recordSets) {
     return;
@@ -222,8 +224,10 @@ export const Records = () => {
             onClick={() => {
               history.push('/Records/List/Local:' + set);
             }}
-            onMouseOver={setHighlightedSet.bind(this, set)}
-            onMouseOut={setHighlightedSet.bind(this, null)}
+            onMouseOver={highlightSet.bind(this, set)}
+            onFocus={highlightSet.bind(this, set)}
+            onMouseOut={unHighlightSet}
+            onBlur={unHighlightSet}
             className={'records_set'}
             // Adjust Opacity of the background-color when hovering
             style={{ backgroundColor: `${recordSets?.[set]?.color}${highlightedSet === set ? 65 : 20}` }}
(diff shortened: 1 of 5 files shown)


plasticviking:
Can you rebase this on to current to make the changeset smaller/easier to review?

app/src/UI/Overlay/Records/Records.tsx
```diff
@@ -211,242 +208,193 @@ export const Records = () => {
...
+            onClick={() => {
+              history.push('/Records/List/Local:' + set);
+            }}
+            onMouseOver={highlightSet.bind(this, set)}
+            onFocus={highlightSet.bind(this, set)}
+            onMouseOut={unHighlightSet}
+            onBlur={unHighlightSet}
+            className={'records_set'}
+            // Adjust Opacity of the background-color when hovering
+            style={{ backgroundColor: `${recordSets?.[set]?.color}${highlightedSet === set ? 65 : 20}` }}
```
plasticviking: better to do this in css / with a class? (or just a :hover pseudo-element).

app/src/UI/Overlay/Records/Records.tsx
```diff
@@ -211,242 +208,193 @@ export const Records = () => {
...
+              <Tooltip
+                classes={{ tooltip: 'toolTip' }}
+                title="Toggle viewing the labels on the map for this layer.  If more than 200 are in the extent, you may need to zoom in to see what you are looking for.  For people on slow computers - it recalculates on drag and zoom so fewer small drags will decrease loading time."
+              >
+                <IconButton
+                  className="records__set__layer_toggle"
+                  onClick={(e) => onClickToggleLabel(set, e)}
+                  color="primary"
+                >
+                  {recordSets?.[set]?.labelToggle ? <LabelIcon /> : <LabelClearIcon />}
```
plasticviking: I think the first elvis is superfluous?

app/src/UI/Overlay/Records/Records.tsx
```diff
@@ -211,242 +208,193 @@ export const Records = () => {
...
+              <Tooltip
+                classes={{ tooltip: 'toolTip' }}
+                title="Toggle viewing the layer on the map, and including these records in the Whats Here search results."
+              >
+                <IconButton
+                  className="records__set__layer_toggle"
+                  onClick={(e) => onClickToggleLayer(set, e)}
+                  color="primary"
+                >
+                  {recordSets?.[set]?.mapToggle ? <LayersIcon /> : <LayersClearIcon />}
```
plasticviking: same

app/src/UI/Overlay/Records/Records.tsx
LocalNewsTV: @plasticviking The background colours are set programmatically based on the position in the colour array, so this is just adding opacity onto the hex values. Otherwise the options are set opacity for the entire container

If we were using emotion/styled we could pass the hex code in as a prop. and `&:hover` would be the ideal
Happy to be wrong though :)

plasticviking (approved):

PR #3514, opened 2024-09-24
[INV-3497] Get bounding box given a filter parameter

LocalNewsTV:
# Overview

> One step towards offline map caching

New API endpoint `/v2/activities/bbox`:
- Using the same filter rules applied to a recordset, this endpoint can be targetted to make a bounding box for all the records contained within it.
- Uses the existing `sanitizeActivityFilterObject()` and `getActivitiesSQLv2()` functionality, reducing maintenance and providing same level of security.

## Testing

Copied Filter payloads from recordset network requests and used them against this new endpoint, Rendered the response objects in an external tool. Results came out as expected, given how bounding boxes function

> [!NOTE]
> This Endpoint is not wired up to anything, I tested manually using the ThunderClient extension in VSC

### Payload example *(Every record made by me)*:

```json
{
	"filterObjects": [
		{
			"limit": 200000,
			"recordSetType": "Activity",
			"selectColumns": [
				"activity_id"
			],
			"tableFilters": [
				{
					"field": "created_by",
					"filter": "EXAMPLE_ID",
					"filterType": "tableFilter",
					"id": "FILTER_ID_EXAMPLE",
					"operator": "CONTAINS",
					"operator2": "AND"
				}
			]
		}
	]
}
```

### API Response:

```json
{
  "bbox": "POLYGON((-127.03320312500054 48.42276700740098,-127.03320312500054 55.69957876911636,-118.78194475016757 55.69957876911636,-118.78194475016757 48.42276700740098,-127.03320312500054 48.42276700740098))"
}
```

### Response displayed using a [MapViewer](https://9revolution9.com/tools/geo/wkt_geojson/):

image

### DIFF
--- a/api/src/paths/recordset-bbox.ts
+++ b/api/src/paths/recordset-bbox.ts
@@ -29,7 +29,13 @@ POST.apiDoc = {
       content: {
         'application/json': {
           schema: {
-            properties: {}
+            type: 'object',
+            properties: {
+              bbox: {
+                type: 'string',
+                description: 'Bounding box for the given filters'
+              }
+            }
           }
         }
       }
@@ -67,7 +73,7 @@ const bboxSql = (cte: SQLStatement): SQLStatement => {
  * @desc Create Bounding box based on the filter properties for a given recordset
  */
 function postHandler(): RequestHandler {
-  return async (req, res, next) => {
+  return async (req, res) => {
     const connection = await getDBConnection();
     if (!connection) {
       return res.status(503).json({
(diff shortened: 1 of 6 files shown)


api/src/paths/recordset-bbox.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+};
+
+/**
+ * @desc modify getActivitiesSQLv2 query to return a bounding box for the recordset
+ * @param cte ActivitiesSQL Query being used as Common Table Expression (CTE)
+ * @returns { SQLStatement } Bounding box SQL
+ */
+const bboxSql = (cte: SQLStatement): SQLStatement => {
```
plasticviking: This does work, but there must be a better way? This is a regression waiting to happen if we change the other query. Maybe we need to refactor some of that query generation code into a more generic form first

PR #3518, opened 2024-09-25
[INV-3498] Initial Action Creators -- Static Class Based

LocalNewsTV:
# Overview

This PR includes the following proposed change(s):

Create static classes for `GeoTracking`, `Alerts`, and `Prompts`
- take object oriented approach in function naming, for `createAction` functions
    - All functions are static, no need for `const Prompt = new Prompt()`
- Small refactoring of functionality in two files (comments added to file diffs flagging them)
- Closes #3498

PR #3522, opened 2024-09-26
3488 initial action creator (Now with overhauled Reducers)

LocalNewsTV:
# Overview

- In addition to the changes that were highlighted in #3518 This overhauls the reducers to use type guards
    - This is branched off of that PR as an addition 
- Added Union Class for all the Modals
- added the redux action `NEW_PROMPT` as a static class member to the `Prompt` class, adding tighter coupling of it to the existing member function. 
- Added missing (unused) clearAll functionality to Prompt class. 

> [!NOTE]
> Since all new Prompts are treated the same in regard to handling, `NEW_PROMPT` still uses string matching, so action type is cast for clarity.
> The alternative was having an identical check for each type of modal that yielded the same result.


app/src/state/reducers/alertsAndPrompts.ts
LocalNewsTV, line 1: Reducer Overhaul

app/src/state/reducers/alertsAndPrompts.ts
```diff
@@ -19,41 +20,26 @@ export function createAlertsAndPromptsReducer(
...
+      } else if (Alerts.deleteAll.match(action)) {
+        draftState.alerts = [];
+      } else if (Prompt.closeOne.match(action)) {
+        draftState.prompts = state.prompts.filter((prompt) => prompt.id !== action.payload.id);
+      } else if (Prompt.closeAll.match(action)) {
+        draftState.prompts = [];
+      } else if (Prompt.NEW_PROMPT.match(action.type)) {
+        const newPrompt: PromptAction = action.payload;
+        const newID = nanoid();
+        draftState.prompts = [...state.prompts, { ...newPrompt, id: newID }];
```
LocalNewsTV: Reducer Overhaul. L39 kept as string, but coupled to class

PR #3555, opened 2024-10-09
Fix Biocontrol suggestions bug, refactor to createActions

### LINKED ISSUE #3556: Biocontrol suggestions showing as `no options`
When creating a new form, Biocontrol options show up as `No options` and required, so users cannot validate

LocalNewsTV:
# Overview

- Fixed an Action string that was mismatched from its variable name. Key was still unique, so was not responsible for any bugs.
- Fixed issue where *new* biocontrol forms would erase the plantToAgentMap due to a mismatched key using a plural noun. 
- Converted Biocontrol functions to actionCreators
- Added null coalescing operator to an action that was causing an error in my iOS simulator when `withLocalToggles` was showing as undefined, now when undefined the map functions will target an empty array. 

- Closes #3556


app/src/state/reducers/activity.ts
```diff
@@ -147,7 +162,7 @@ function createActivityReducer(): (ActivityState, AnyAction) => ActivityState {
               loading: false,
               saved_activity_hash: null,
               biocontrol: {
-                plantToAgentMap: draftState.biocontrol.plantsToAgentMap ?? [],
+                plantToAgentMap: draftState.biocontrol.plantToAgentMap ?? [],
```
LocalNewsTV, line 165: Caused the bug with Biocontrol fields

app/src/state/reducers/map.ts
```diff
@@ -897,15 +897,16 @@ function createMapReducer(configuration: AppConfig): (MapState, AnyAction) => Ma
...
+            const withLocalToggles =
+              action.payload.data?.map((incomingItem) => {
+                const returnVal = { ...incomingItem };
+                const existingToggleVal = draftState.serverBoundaries.find((oldItem) => {
+                  oldItem.id === incomingItem;
+                })?.toggle;
+                returnVal.toggle =
+                  existingToggleVal !== null && existingToggleVal !== undefined ? existingToggleVal : false;
+                return returnVal;
+              }) ?? [];
```
LocalNewsTV, line 909: Fixes a niche issue where I experienced a "Cannot map undefined" on serverBoundaries in iOS.
Nobody has previously mentioned this before.

app/src/state/reducers/activity.ts
plasticviking: nice

plasticviking (approved):

Issue #3428, opened 2024-08-19
Finish Manual UTM Entry feature


LocalNewsTV:
# Describe the task
> If the task will only make sense to a dev, add a :zap: to the title.

There is a feature in `Form.tsx` to create a Manual UTM entry, but after getting the information from the user this feature stops, and the geo object created by the user doesn't get used anywhere

## Acceptance Criteria

- [ ] Information from Manual UTM Entry is used in the form

## Additional Context:
- Need understanding in what the complete workflow will need to be for this feature
    - If it can feed into an existing workflow, or needs a unique redux set to occur

PR #3561, opened 2024-10-10
[INV-3528] 'Suggested' ActionCreators Converted Over

LocalNewsTV:
# Overview
> A Portion of the work being done for #3428

1. Draftstate for Activity is now typed, finding errors 
    - Stubbed the keys that were flagged as non-existent from this change
2. Created type for SuggestedTreatmentId
3. Finished converting `Suggestions` type actions into Action Creators
4. Tidied up typing, reducing warnings.

## Unused Keys

> Following keys were unused, and deleted.
 
- ACTIVITY_GET_SUGGESTED_PERSONS_REQUEST_OFFLINE 
- ACTIVITY_GET_SUGGESTED_TREATMENT_IDS_REQUEST_OFFLINE

## Testing
- Created form, Results from Dev / Local matched on Linked treatment Ids
- Created form, If you empty the string containing your name, it still repopulates it with your name


app/src/state/reducers/activity.ts
```diff
@@ -24,37 +22,42 @@ import GeoShapes from 'constants/geoShapes';
 import { CURRENT_MIGRATION_VERSION, MIGRATION_VERSION_KEY } from 'constants/offline_state_version';
 import GeoTracking from 'state/actions/geotracking/GeoTracking';
 import Activity from 'state/actions/activity/Activity';
+import SuggestedTreatmentId from 'interfaces/SuggestedTreatmentId';
 
 export interface ActivityState {
   [MIGRATION_VERSION_KEY]: number;
   activity: any;
+  activeActivity: string | null;
+  activityErrors: any[];
```
LocalNewsTV, line 32: These two keys were being added/modified in the reducer, but were not present in the interface

app/src/state/reducers/activity.ts
```diff
@@ -79,9 +82,9 @@ const initialState: ActivityState = {
   uiSchema: undefined
 };
 
-function createActivityReducer(): (ActivityState, AnyAction) => ActivityState {
+function createActivityReducer(): (ActivityState: ActivityState, AnyAction) => ActivityState {
   return (state = initialState, action) => {
-    return createNextState(state, (draftState) => {
+    return createNextState(state, (draftState: ActivityState) => {
```
LocalNewsTV: Matching the type to `draftState` here will now flag missing/mismatched keys in the reducer.
Adding this declaration flagged the keys being added that were missing in the interface.

plasticviking: I think it should be `Draft<ActivityState>`, since we're using `immer`

LocalNewsTV: @plasticviking  This is now corrected

PR #3565, opened 2024-10-16
[INV-3563] Biocontrol FIltering Fix, New widget

### LINKED ISSUE #3563: Biocontrol plant/agent filtering doesn't work for multiple different plants
If you add a second plant, the only agent options will still be filtered by the first plant, so you can't pick the correct agent

LocalNewsTV:
# Overview

- Created Widget for managing the filters of the Biocontrol Agents
    - Work largely based off of existing `SingleSelectAutocomplete.tsx`
    - Widget utilizes the 'last updated' component to update when the adjacent plant code updates through regex matching/swapping. Given there is four different combinations of keys that can be used, depending on the form using it, this was chosen as the preferred method.
- Removed Redux actions associated with editing the filters for Biocontrol agents, as this is now handled in the widget itself
- Converted use of `biological_agent_code` to use the new widget
- Closes #3563

## Testing
- Created and submitted Two forms, each with two separate plant/agent combos, filtering works as expected
- Created and Submitted (as draft) remaining two forms, each with two separate plant/agent combos, filtering works as expected


plasticviking (approved):
lgtm

PR #3599, opened 2024-10-22
[INV-3574] Bugfix** GeoTracker crash when location not enabled

### LINKED ISSUE #3574: Mobile: trying to use geometry tracker with location tracker off causes crash
After selecting linestring or polygon and clicking ok. TypeError: null is not an object (evaluating 'coords.hasOwnProperty').

I don't know what the best solution is for this. It would be nice if clicking the geometry tracker turned on location tracking automatically because I can see people forgetting to do it themselves. But I don't know if the gps accuracy will be bad for the first few seconds and therefor mess up the geometry...we can talk about this.

An easy fix would probably be just popping an error/suggestion in the top right that says you need to turn on location tracking to use the geometry tracker.

LocalNewsTV:
# Overview

- Convert Message to constant using IIFE syntax
- adding Elvis to `.hasOwnProperty()` call (this was the cause of error)
- Added Alert messages for Location settings being offline to better communicate to user
- Added termination of GeoTracking when coordinates unavailable at start.
- Closes #3574

### DIFF
--- a/app/src/state/sagas/activity.ts
+++ b/app/src/state/sagas/activity.ts
@@ -231,7 +231,7 @@ function* handle_MAP_TOGGLE_TRACK_ME_DRAW_GEO_START(action) {
       message = mappingAlertMessages.trackingStarted;
   }
 
-  const initCoords = coords.hasOwnProperty('long') ? [[coords.long, coords.lat]] : [];
+  const initCoords = coords?.hasOwnProperty('long') ? [[coords.long, coords.lat]] : [];
   const initGeo = {
     type: 'Feature',
     properties: {},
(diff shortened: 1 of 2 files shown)


app/src/state/sagas/activity.ts
```diff
@@ -215,33 +215,37 @@ function* handle_ACTIVITY_BUILD_SCHEMA_FOR_FORM_REQUEST(action) {
...
-      message = mappingAlertMessages.trackingStartedLineString;
-      break;
-    case GeoShapes.Polygon:
-      message = mappingAlertMessages.trackingStartedPolygon;
-      break;
-    default:
-      message = mappingAlertMessages.trackingStarted;
-  }
-
-  const initCoords = coords.hasOwnProperty('long') ? [[coords.long, coords.lat]] : [];
```
LocalNewsTV, line 234: Line is cause of bug due to coords being null.

plasticviking (approved):

PR #3636, opened 2024-10-30
[INV-3628] Block starting multiple downloads of the same cache

### LINKED ISSUE #3628: Mobile: Can start download for cached area multiple times

LocalNewsTV:
# Overview

- Erase Name after starting download to prevent starting multiple downloads
- Move whitespace into the conditional render, since it is trailing when the render == false
- Wipe Geography after two seconds, forcing render to 'No shape selected' default
    - Deemed the delay would prevent user confusion from it immediately flipping
- Separated download functionality into named one
- Closes #3628

### DIFF
--- a/app/src/UI/Overlay/TileCache/TileCacheCreationPanel.tsx
+++ b/app/src/UI/Overlay/TileCache/TileCacheCreationPanel.tsx
@@ -9,6 +9,20 @@ import { useTileSizeThresholds } from './tileSizeHook';
 import { APPROX_SIZE_PER_TILE, AVAILABLE_ZOOMS } from './constants';
 
 const TileCacheCreationPanel = () => {
+  const handleDownload = () => {
+    if (!drawnShape) return;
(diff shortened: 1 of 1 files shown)


app/src/UI/Overlay/TileCache/TileCacheCreationPanel.tsx
```diff
@@ -9,6 +9,20 @@ import { useTileSizeThresholds } from './tileSizeHook';
 import { APPROX_SIZE_PER_TILE, AVAILABLE_ZOOMS } from './constants';
 
 const TileCacheCreationPanel = () => {
+  const handleDownload = () => {
+    if (!drawnShape) return;
```
LocalNewsTV, line 13: Since the function sits at the top of the component and no longer inline with the render, the `drawnShape` check has to be done to assert its not null

plasticviking (approved):

Issue #3637, opened 2024-10-31
Make App Mobile Friendly


LocalNewsTV:
# Describe the task
> If the task will only make sense to a dev, add a :zap: to the title.

> [!NOTE]
> Its primarily to make the app `usable` on a Cellphone. Look and feel unlikely to change for most sections.

## Value Add
- Wider range of user options
- Can check into the Web/Mobile app while on the go without need for Tablet / PC

## Acceptance Criteria

- [x] Widen the Header options to use more screen space
- [x] Map options overflow out, cannot change Map tiles, or see last button
- [x] Recordset Lists squished in `/Records`
- [x] #3513
- [x] Weird border-right-like property appears in Home Text
- [x] Top section of forms overflows out of (RJSF doesn't)
- [x] Camera button partially obstructed by footer
- [x] Tooltips overflow out of view
- [ ] ~Batch uploads rows overflow~
- [ ] ~Batch Upload Page overflows~
- [ ] ~Download Batch CSVs has major inconsistencies~
- [x] Admin Panel Email templates get Squished
- [x] Alerts overflow from Mobile
- [x] #3503 
- [x] Offline Sync Status Menu
- [x] Logo Tiny

PR #3652, opened 2024-11-05
[INV-3637] PARTIAL** Mobile Friendly changes

LocalNewsTV:
# Overview

> This is partial work towards the completion of #3637

> Viewport was set to Galaxy S20

1. Widen the header to better use screen space
    - Flip Network icons horizontal instead of stacked. They're too large to fit in pre-existing header size. 
    - switch outer `div` to `header` given its a header
    - CSS cleanup
2. Refactor Map Buttons to not overflow out and disappear
    - CSS Adjustments only
3. Recordset lists
    - Break /Records into sub components
    - Split into two main subcomponents
    - In Mobile widths, stack them to avoid compression
    - Editing a recordset name doesn't toggle them all into edit mode
    - Reduce logic complexity in what is supposed to render and when

## Pictures
1. Widening Header

New

image

Old

image

2. Map Buttons

New

image

Old

> This as far left as you can scroll, given context, buttons become unusable, rightmost button is eclipsed by map details

image

3. Recordsets

Old

image
image

New 

image

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -1464,7 +1464,8 @@
     "node_modules/@types/geojson": {
       "version": "7946.0.14",
       "resolved": "https://registry.npmjs.org/@types/geojson/-/geojson-7946.0.14.tgz",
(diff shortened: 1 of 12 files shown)


LocalNewsTV:
Both Sonarcloud issues are pre-existing prior to this work

PR #3673, opened 2024-11-14
[INV-3672] Linked ID showing improperly

### LINKED ISSUE #3672: Short ID's showing improperly
- User Sees incorrect short Id's when trying to create a monitoring record
- Short id's are linked across province

Estimated Users Affected: <= 5

LocalNewsTV:
# Overview

- Initial bug was New activities firing LinkedID calls when creating a new record and a geography had not been set
    - Blocked calls when shape doesn't exist.
- Moved component into own widget, separating from the existing catch-all `SingleSelectAutoComplete.tsx`
    - Removed object drilling that occurs with Linked_Id state in the reduce
    - Create `LinkedIdSelectAutoComplete.tsx` to handle the state of the widget
- Correct max-height attribute for `AgentSelectAutoComplete` for tidiness.
- Correct text in Alert
- Closes #3672

### DIFF
--- a/app/src/rjsf/widgets/AgentSelectAutoComplete.tsx
+++ b/app/src/rjsf/widgets/AgentSelectAutoComplete.tsx
@@ -86,6 +86,11 @@ const AgentSelectAutoComplete = (props: WidgetProps) => {
       onLoad={() => {
         props.onChange(value);
       }}
+      SelectProps={{
+        MenuProps: {
+          sx: { height: '300px' }
+        }
+      }}
     >
       {filteredOptions.map((entry) => (
         <MenuItem key={entry.value} value={entry.value}>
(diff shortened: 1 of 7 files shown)


app/src/rjsf/widgets/AgentSelectAutoComplete.tsx
```diff
@@ -86,6 +86,11 @@ const AgentSelectAutoComplete = (props: WidgetProps) => {
       onLoad={() => {
         props.onChange(value);
       }}
+      SelectProps={{
+        MenuProps: {
+          sx: { height: '300px' }
+        }
+      }}
```
LocalNewsTV, line 93: Recommended, due to MUI instability

PR #3696, opened 2024-11-18
[INV-3662] Create Error Handler, move middleware to named files, create common Database caller

### LINKED ISSUE #3662: :zap: Common Error Handler in API
# Describe the task

Create a error handling middleware to gracefully handle errors across API Calls, simplifying our endpoints.

## Value Add
- Reduced Complexity
- More DRY

## Acceptance Criteria

- [x] Error handler created to gracefully handle errors

LocalNewsTV:
# Overview

- Create modular error handler as a catch all
    - leverages the `Request` object to track filepaths instead of manual entry 
    - utilizes the Logger 
    - Overridable status code when used with the `CustomError` class
    - Closes #3662
- Moved middleware into named files for modularity
- Created common Database caller
    - Handles setup/teardown of database connections to be more DRY 

- While the global error handler will function right away, the database caller will be implemented later in sprint


api/src/utils/commonDbRequest.ts
```diff
@@ -0,0 +1,25 @@
+import { getDBConnection } from 'database/db';
```
plasticviking: I'd prefer to solve this problem in a more OOP way, by creating a base class from which we derive our action handlers.

LocalNewsTV: Removed existing work, will implement as a new PR at later point.

PR #3699, opened 2024-11-20
[INV-3682][INV-3509] Recordset Filtering bugs

### LINKED ISSUE #3509: Mobile: Crash when filtering recordset
While online and filtering InvasivesBC recordset for invasive plant the app crashed. Happened multiple times, no error was shown, the app just restarts

Note: typing in the filter box is also slow for some reason

### LINKED ISSUE #3682: Mobile: Multiple filters on recordsets slows loading
This may be a problem on web also, in which case I can try to optimize the query. I think there is definitely a use case for people using 5+ filters when trying to cache a recordset.

LocalNewsTV:
# Overview

> [!IMPORTANT]
> Given a simulator, the bugs were not replicable and did not throw any logs that might attribute to what caused the bugs. Assumption becomes large spikes of data in Ram getting parsed asynchronously causing failures. It is recommended to test this on a physical device before confirming the completion of #3509

## Backend
- Fixed bug where filters on `Elevation` did not work
- Removed a duplicate where clause in filter statement

## Frontend
- Split `Recordset.tsx` into separate files per component, creating a lot of diffs in this pr
- Added 0.75s timeout to updates on Filter input fields, so users get updated lists when they finish typing instead of sending / receiving `n` requests where `n` is key presses.
    - This should help reduce spikes in the API/DB Memory caused by rapid retrieval and sending of larger datasets that get
    - This should also avoid bogging down resources in user devices from parsing massive queries / updating state
- Move `RECORDSET_REMOVE/DELETE_FILTER` actions to actionCreators

## General
- Misc. fixes that **Removed >= 25 pre-existing Code smells**
- Closes #3682


app/src/UI/Overlay/Records/RecordSet/Filter.tsx
```diff
@@ -0,0 +1,289 @@
...
+  };
+
+  /**
+   * @desc Change Handler for text input filter.
+   *       Sets/refreshes timeouts so updates won't fire until user finishes typing.
+   */
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+    clearTimeout(timeoutRef.current as NodeJS.Timeout);
+    timeoutRef.current = setTimeout(() => {
```
plasticviking: I'd suggest just using lodash throttle

PR #3707, opened 2024-11-22
Public Map component 

### LINKED ISSUE #3690: Make public map new react-components version
- [x] Public map is react component version
- [x] Topo, Sat (HD & SD) all work
- [x] Zoom controls work
- [x] Public pmtile layers work

meghna0593:
# Overview

- Added new library to render maps
- Added public pmtiles layer across all three base maps, ensuring mobile compatibility 
- Added navigation and scale controls for the public layer
- Added a record set layer (WIP), but disabled in this PR

## Type of change

- [x] Refactoring
- [ ] Version change


api/package-lock.json
LocalNewsTV, line 1: Was deleting the backend's package-lock intentional? All the changes in the PR are for the frontend

app/src/UI/App.tsx
```diff
@@ -227,6 +229,7 @@ const OverlayContentMemo = () => {
 
 const App: React.FC = () => {
   const authInitiated = useSelector((state) => state.Auth.initialized);
+  const authenticated = useSelector((state: any) => state.Auth.authenticated);
```
LocalNewsTV, line 232: The useSelector imported from 'utils/use_selector' is typed.  There's no need for `type: any` here.

app/src/UI/Map2/Map3.tsx
```diff
@@ -0,0 +1,246 @@
...
+              "https://server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}"
+            ]}
+            tileSize={256}
+          >
+            <Layer
+              id='wms-layer'
+              type='raster'
+              source='esri-sat-layer-hd'
+            />
+          </Source> */}
```
LocalNewsTV, line 225: Leftover comments

app/src/UI/Map2/RecordSetLayers3.tsx
```diff
@@ -0,0 +1,94 @@
+import { Source, Layer } from 'react-map-gl/maplibre';
+import { useSelector } from 'react-redux';
```
LocalNewsTV: Change this to the import from 'utils/use_selector' so that it's strongly typed. then correct lines 41/42

app/src/UI/Map2/Map3.tsx
```diff
@@ -0,0 +1,246 @@
+import * as React from 'react';
+import { useSelector } from 'react-redux';
```
LocalNewsTV: Same as this one. `utils/use_selector`

app/src/UI/Map2/Map3.tsx
```diff
@@ -0,0 +1,246 @@
+import * as React from 'react';
```
LocalNewsTV: There's an entire react import here, but the values being used are destructured on line 3

LocalNewsTV (changes requested):
It looks like you don't have our linter settings configured correctly.
There is a lot of formatting issues across the PR that will be easily resolvable once the linter is setup

We do have Prettier configs setup in both the front and back ends, so when prettier/eslint are setup correctly, they should run automatically, without needing to worry about this again

The New files added are missing [Blank lines at the end](https://stackoverflow.com/questions/77201426/what-does-a-red-icon-with-a-hyphen-mean-in-a-github-pr)

Assuming you're using VSC, I recommend enabling these options in your settings
image

If you aren't, your IDE will likely have some similarly named options

app/src/UI/Map2/Map3.tsx
meghna0593: This was intentional, to revert to component based, if need be.

app/src/UI/Map2/Map3.tsx
meghna0593: good catch!

api/package-lock.json
meghna0593: yes

PR #3722, opened 2024-11-27
[INV-3715] Use Network to determine online status v1

### LINKED ISSUE #3715: Utilize Health Check to determine if Online/Offline
# Describe the task
Create common function to target the APIs `misc/version` endpoint to determine health connectivity

## Value Add
- Helps start the application offline when in Mobile, avoiding functions that hang at the start

## Acceptance Criteria
- [x] Given I have no internet connection when I start the application, I am defaulted to Offline mode
- [x] If I fail the health check, move my status to offline
- [x] If I click to go online, the health endpoint is checked, if I fail, I remain offline with a notification

## Additional Context
- We're already using the misc/version endpoint to determine container health, hence why it was chosen

LocalNewsTV:
# Overview

Handling These Network Scenarios
- Initial Network State
- Manual User Disconnect/Reconnect
- Rolling API checks to maintain current network status
- Automatic re-attempts to connect if disconnected

**V2 Iteration can use Websockets to abstract all this work.**

## Changes:

> All changes affect Mobile Only

- Closes #3715 
- Do one test connection against API to determine initial Network state for Online/Offline
    - The multi-checks before fail for the normal loop resolves too slow to be effective at start 
- Begin Network loop when User attempts to come online
- Cancel Network loop when user goes offline
- Monitor Network connection through API calls, similarly to how auth tokens are updated.
- If User determined offline, communicate to user, then attempt to reconnect in background. Alert user if these reconnect attempts fail

### DIFF
--- a/app/src/UI/Header/Header.tsx
+++ b/app/src/UI/Header/Header.tsx
@@ -377,7 +377,7 @@ const LoginOrOutMemo = React.memo(() => {
 
 const NetworkStateControl: React.FC = () => {
   const handleNetworkStateChange = () => {
-    dispatch(connected ? NetworkActions.offline() : NetworkActions.checkMobileNetworkStatus());
+    dispatch(connected ? NetworkActions.offline() : NetworkActions.manualReconnect());
   };
   const { connected } = useSelector((state) => state.Network);
   const dispatch = useDispatch();
--- a/app/src/UI/Header/Header.tsx
+++ b/app/src/UI/Header/Header.tsx
@@ -377,7 +377,7 @@ const LoginOrOutMemo = React.memo(() => {
 
 const NetworkStateControl: React.FC = () => {
   const handleNetworkStateChange = () => {
-    dispatch(connected ? NetworkActions.offline() : NetworkActions.online());
+    dispatch(connected ? NetworkActions.offline() : NetworkActions.checkMobileNetworkStatus());
   };
   const { connected } = useSelector((state) => state.Network);
   const dispatch = useDispatch();
--- a/app/src/state/sagas/network.ts
+++ b/app/src/state/sagas/network.ts
@@ -1,7 +1,11 @@
 import networkAlertMessages from 'constants/alerts/networkAlerts';
+import { HEALTH_ENDPOINT } from 'constants/misc';
 import { all, put, select, takeEvery } from 'redux-saga/effects';
 import Alerts from 'state/actions/alerts/Alerts';
 import NetworkActions from 'state/actions/network/NetworkActions';
+import { MOBILE } from 'state/build-time-config';
+import { selectConfiguration } from 'state/reducers/configuration';
+import { selectNetworkConnected } from 'state/reducers/network';
 import { OfflineActivitySyncState, selectOfflineActivity } from 'state/reducers/offlineActivity';
 
 function* handle_NETWORK_GO_OFFLINE() {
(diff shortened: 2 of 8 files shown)


meghna0593 (dismissed):

app/src/state/sagas/network.ts
```diff
@@ -20,10 +24,33 @@ function* handle_NETWORK_GO_ONLINE() {
...
+
+  const networkCheckPassed = yield fetch(configuration.API_BASE + HEALTH_ENDPOINT)
+    .then((res) => res.status === 200)
+    .catch(() => false);
+
+  if (!networkCheckPassed && !currentOnlineStatus) {
+    yield put(Alerts.create(networkAlertMessages.attemptToReconnectFailed));
+  } else if (!networkCheckPassed) {
+    yield put(NetworkActions.userLostConnection());
+    yield put(Alerts.create(networkAlertMessages.userLostConnection));
```
plasticviking: I'd rather see the alerts decoupled from the check function (by `take`ing userLostConnection for example).

app/src/UI/Header/Header.tsx
```diff
@@ -377,7 +377,7 @@ const LoginOrOutMemo = React.memo(() => {
 
 const NetworkStateControl: React.FC = () => {
   const handleNetworkStateChange = () => {
-    dispatch(connected ? NetworkActions.offline() : NetworkActions.online());
+    dispatch(connected ? NetworkActions.offline() : NetworkActions.checkMobileNetworkStatus());
```
plasticviking: I don't love putting this check here. This control was mostly intended as a placeholder. Or, it could be here since you are effectively not trusting the user to know better, but I think in that case it should also be dispatched on app startup (if online).

app/src/state/sagas/network.ts
```diff
@@ -1,13 +1,103 @@
...
+      }
+    } while (!canConnect && attempts < MAX_ATTEMPTS);
+
+    if (!canConnect) {
+      yield put(NetworkActions.userLostConnection());
+      return;
+    }
+    yield delay(SECONDS_BETWEEN_CHECKS * 1000);
+  }
+}
```
meghna0593, line 62: a single do-while loop should suffice? You have about 5 attempts going on here with a delay of 5 seconds. You can  increase the MAX_ATTEMPTS and remove the outer while loop or increase SECONDS_BETWEEN_ATTEMPTS.

app/src/state/sagas/network.ts
```diff
@@ -1,13 +1,103 @@
...
+  } else {
+    yield put(NetworkActions.offline());
+  }
+}
+/**
+ * @desc Handler for User manually going offline. Fires a cancellation event to stop rolling api checks.
+ */
 function* handle_NETWORK_GO_OFFLINE() {
   yield put(Alerts.create(networkAlertMessages.userWentOffline));
+  yield put(NetworkActions.checkMobileNetworkStatus(true));
```
meghna0593, line 94: Why do we have this here? It looks like sending a `cancel` payload set to `true` in handle_CHECK_MOBILE_NETWORK_STATUS does nothing

app/src/state/sagas/network.ts
LocalNewsTV: @meghna0593 
the `while(true)` loop is always pinging the API to ensure continued connection unless otherwise cancelled (by a user choosing to go offline, or being disconnected and unable to reconnect)

If that do-while fetch request fails and we are UNABLE to reach our API, we don't want to immediately panic, so we send a few more requests at a faster interval to see if anything will come back. If we still can't reach the API we will trigger Offline mode. 

While in offline mode, we will sporadically try a few more attempts to come back online.

In the end this is just v1 functionality, it will be succeeded by WebSockets or some similar network equivalent

app/src/state/sagas/network.ts
LocalNewsTV: @meghna0593 
because we're handling this action with `takeLatest` when a new call is made to this saga, redux will automatically attempt to cancel the old one. The new one will exit early due to the `cancel` flag being an early return

app/src/state/sagas/network.ts
LocalNewsTV: Originally this PR was VERY different than it is now. 
The previous code you reviewed still had a `handle_CHECK_MOBILE_NETWORK_STATUS` function, but the logic for it is black and white compared to the logic I explained it in call :) 
Sorry for any confusion !

meghna0593: if this was intended to be on an infinite loop when `canConnect` is `true`,  then do you think this could be blocking in nature? Maybe making this a background task is worth looking at? @plasticviking thoughts?

PR #3726, opened 2024-11-29
Mobile: Activity button on Offline Sync Status Dialog box doesn't load activity

### LINKED ISSUE #3684: Mobile: Load activity button on offline sync status menu doesn't work
The floppy disk icon on the offline sync status menu does not take you to the activity when clicked.

meghna0593:
# Overview
Resolves the issue where the dialog box does not close when attempting to load offline activity.

## Type of change

- [ ] Version change

### DIFF
--- a/app/src/UI/OfflineDataSync/OfflineDataSyncTable.tsx
+++ b/app/src/UI/OfflineDataSync/OfflineDataSyncTable.tsx
@@ -3,7 +3,11 @@ import { Button, IconButton, LinearProgress } from '@mui/material';
 import { OfflineActivityRecord, selectOfflineActivity } from 'state/reducers/offlineActivity';
 import { useSelector } from 'utils/use_selector';
 import { useDispatch } from 'react-redux';
-import { ACTIVITY_OFFLINE_DELETE_ITEM, ACTIVITY_RUN_OFFLINE_SYNC } from 'state/actions';
+import {
+  ACTIVITY_OFFLINE_DELETE_ITEM,
+  ACTIVITY_OFFLINE_SYNC_DIALOG_SET_STATE,
+  ACTIVITY_RUN_OFFLINE_SYNC
+} from 'state/actions';
 import Delete from '@mui/icons-material/Delete';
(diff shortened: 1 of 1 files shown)


app/src/UI/OfflineDataSync/OfflineDataSyncTable.tsx
```diff
@@ -63,8 +66,8 @@ export const OfflineDataSyncTable = () => {
                         disabled={!(workingOffline || authenticated)}
                         color="primary"
                         onClick={() => {
-                          dispatch(Activity.getLocal(key));
```
LocalNewsTV, line 66: Nice catch on the duplicate run of getLocal

LocalNewsTV (commented):
I'm not sure if we can confidently call this enough to close the ticket.

One outlying issue is that we're not caching the API Docs (because they are massive) so if we start from an offline context we can't actually load the offline activities due to needing the apidocs.

If we don't have the apiDocs, the form page appears but it just says `loading` and will never terminate. A console log of Record.tsx l:63 
`if (activity_ID && apiDocsWithSelectOptions && apiDocsWithViewOptions && loading === false)` shows that both apiDoc options are null

We may just want to verify if the issue is that we're getting `loading` or if the issue was strictly navigating. On mobile users can't see the URL they're on :)

meghna0593:
> I'm not sure if we can confidently call this enough to close the ticket.
> 
> One outlying issue is that we're not caching the API Docs (because they are massive) so if we start from an offline context we can't actually load the offline activities due to needing the apidocs.
> 
> If we don't have the apiDocs, the form page appears but it just says `loading` and will never terminate. A console log of Record.tsx l:63 `if (activity_ID && apiDocsWithSelectOptions && apiDocsWithViewOptions && loading === false)` shows that both apiDoc options are null
> 
> We may just want to verify if the issue is that we're getting `loading` or if the issue was strictly navigating. On mobile users can't see the URL they're on :)

Issue was just with navigating to activities on clicking the button and the dialog box persisting

PR #3736, opened 2024-12-05
WIP: BCGW layer doesn't work for Mobile

meghna0593:
# Overview

- Proxy to hit the openmaps endpoint to avoid cors issue
- Changes on the frontend to cater to this api

## Type of change

- [ ] Version change

### DIFF
--- a/api/src/paths/proxy/openmaps.ts
+++ b/api/src/paths/proxy/openmaps.ts
@@ -1,0 +1,88 @@
+import { RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { ALL_ROLES, SECURITY_ON } from 'constants/misc';
+import { getLogger } from 'utils/logger';
+import { InvasivesRequest } from 'utils/auth-utils';
+const https = require('https');
+
+const defaultLog = getLogger('proxy');
+
+const GET: Operation = [getOpenMapsWMSTiles()];
+
+GET.apiDoc = {
+  description: 'Proxy requests to a BCGW WMS endpoint',
+  tags: ['proxy'],
+  parameters: [
+    {
+      name: 'url',
+      in: 'query',
+      required: true,
+      type: 'string',
+      description: 'The target WMS URL to proxy'
+    }
+  ],
+  responses: {
+    200: {
+      description: 'Successful response',
+      schema: {
+        type: 'string'
+      }
+    },
+    400: {
+      description: 'Invalid request'
+    },
+    500: {
+      description: 'Internal server error'
+    },
+    401: {
+      $ref: '#/components/responses/401'
+    },
+    503: {
+      $ref: '#/components/responses/503'
+    },
+    default: {
+      $ref: '#/components/responses/default'
+    }
+  }
+};
+
+/**
+ * Proxy request to Openmaps from mobile, redirect the response back to the client
+ *
+ * @return {RequestHandler}
+ */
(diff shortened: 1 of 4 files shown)


api/src/paths/proxy/openmaps.ts
```diff
@@ -0,0 +1,88 @@
...
+        message: 'Missing URL parameter',
+        request: req.body,
+        namespace: 'openmaps/url={url}',
+        code: 400
+      });
+    }
+
+    https
+      .get(targetUrl, (apiRes) => {
+        if (apiRes.statusCode !== 200) {
```
LocalNewsTV, line 69: Should we not use `status.ok` instead of `!== 200` for instances where the status is 304? 
image

api/src/paths/proxy/openmaps.ts
```diff
@@ -0,0 +1,88 @@
...
+        res.setHeader('Content-Type', apiRes.headers['content-type']);
+        res.setHeader('Cache-Control', apiRes.headers['cache-control'] || 'no-cache');
+
+        // Pipe the API response directly to the client
+        apiRes.pipe(res);
+      })
+      .on('error', (err) => {
+        console.error('Request error:', err);
+        res.status(500).send('Internal Server Error');
+      });
```
LocalNewsTV, line 84: We have a Middleware for global error handling that adds a little more description (namespace/request. :)
 It will just catch the error, so you don't need this block

app/src/UI/Map2/LayerPicker/LpLayers/LpLayers.tsx
```diff
@@ -38,14 +38,15 @@ const LpLayers = () => {
   const WmsLayers = useSelector((state) => state.Map?.simplePickerLayers2);
   const KmlLayers = useSelector((state) => state.Map?.serverBoundaries);
   const drawnLayers = useSelector((state) => state.Map?.clientBoundaries);
+  const isAuth = useSelector((state: any) => state.Auth?.authenticated);
```
LocalNewsTV, line 41: Don't need `:any` because the `utils` useSelector  is typed!
We also don't need to check for Auth, the parent `LayerPicker` has a null render when unauthenticated :)

app/src/UI/Map2/helpers/map-init.ts
```diff
@@ -137,6 +137,16 @@ export const mapInit = (options: MapInitOptions) => {
           }
         };
       }
+
+      // const api_local = 'http://localhost:3002';
```
LocalNewsTV, line 141: Leftover comment

api/src/paths/proxy/openmaps.ts
LocalNewsTV, line 1: SonarCloud flagged some unused variables in this file. 
image

LocalNewsTV (commented):
Just a few minor things
Think this would need a proper test on a physical device ( that I don't have ) to properly sign off on :)

api/src/paths/proxy/openmaps.ts
meghna0593: This was kind of a POC, wanted Rob's input :) I'll format the code to what we have in the app now

meghna0593:
> Just a few minor things Think this would need a proper test on a physical device ( that I don't have ) to properly sign off on :)

Good point, I'll ask Brennan to test it out :D

PR #3738, opened 2024-12-11
[INV-3685][INV-3732][INV-3678] Show Map Markers with Cached Recordsets, hide default recordsets from caching

### LINKED ISSUE #3678: Mobile: Allow caching of default recordsets, or remove caching buttons from default recordsets
I like the idea of being able to cache the default recordsets and having them retain the default colours. If this isn't doable the caching buttons should be removed from the three default recordsets (My Drafts, All InvasivesBC Activities, and All IAPP Records).

### LINKED ISSUE #3685: Mobile: Cached recordsets not visible on map
They work fine when online, but if you have them disabled before going offline and then try to turn them on while offline they will not render.

### LINKED ISSUE #3732: Mobile: disable caching of My Drafts recordset
I don't think there is a use case for being able to do this, it probably just confuses users. Related: should there be a prompt for users to direct them to the add new ibc/iapp layer buttons to cache records?

LocalNewsTV:
# Overview

## [INV-3685]

- Cached recordsets now show on Map when user is offline, Label functionality also enabled
    - Cached Recordsets now contain GeoJSON column on iOS
    - cachedMetadata (redux state) now holds a persisted FeatureCollection object to avoid costly recalculations of filtering the local db to recreate layers

- When using `Offline User` Force them into offline mode
    - Previously offline users couldn't have the correct map context, and map layers would not render, this starts to bridge that gap.
    - Forcing user offline lets us access greater map context, and skips a bunch of unauthorized API calls.
    - `Map.tsx` modified to differentiate between an `authenticated` user from a `loggedIn` user.

- Added functionality to strip layers off the map when going from online -> offline context to populate Caches
- Reworked `recordset-layers.tsx` to be more DRY by turning layer declarations into modifiable functions, promoting reusability
    - Lifted some lambda functions into their own, for re use
- Closes #3685

## [INV-3732][INV-3678] Disable `Default` Recordsets from being eligible for Caching
 
- Closes #3732 
- Closes #3678

## Misc

- Add Refresh button to the error page (mobile users had to Force restart the app or clear their cache. Refresh was not an option)
- Cleaned up several SonarCloud highlighted code smells
    - Things like using `forEach` instead of `map`, simplifying logic flows


app/src/UI/Map2/Map.tsx
```diff
@@ -51,7 +52,7 @@ export const Map = ({ children }) => {
 
   const [draw, setDraw] = useState(null);
   const [mapReady, setMapReady] = useState(false);
-
+  const [loggedIn, setLoggedIn] = useState<boolean>(false);
```
plasticviking: should we put this in the redux store -- something like `loggedInOrWorkingOffline`. We probably use this in multiple places.

app/src/utils/record-cache/sqlite-cache.ts
```diff
@@ -206,33 +221,84 @@ class SQLiteRecordCacheService extends RecordCacheService {
...
+    if (this.cacheDB == null) {
+      throw new Error('cache not available');
+    }
+    const centroidArr: any[] = [];
+    const geoJsonArr: any[] = [];
+    const results = await this.cacheDB?.query(
+      // language=SQLite
+      `SELECT GEOJSON, SHORT_ID
+       FROM CACHED_RECORDS
+       WHERE ID IN (${ids.map(() => '?').join(', ')})
```
plasticviking, line 245: slick

PR #3762, opened 2024-12-24
[INV-3742][INV-3733][INV-3747][INV-3746] IAPP Records can Be cached and Used when Offline

### LINKED ISSUE #3733: Mobile: "New Recordset - Activity" text should be wiped automatically when naming custom recordset
If you add a new ibc/iapp layer of records and click the edit icon to edit the name, you have to manually delete the text. It should just remove it automatically so the user can type the name they want.

### LINKED ISSUE #3742: Mobile: Ability to cache iapp recordset
Need to be able to cache IAPP recordset and view it on the map while offline. Basically all the same requirements as the InvasivesBC recordset.

### LINKED ISSUE #3746: Mobile: Show Cached IAPP Records On Map when Offline
- Same functionality as Activity's, expanded to cover IAPP Records

### LINKED ISSUE #3747: Mobile: Cached IAPP Records work in `Whats Here`
Same functionality as cached Activities using WhatsHere , expanded to cover IAPP Records

LocalNewsTV:
# Overview

> [!TIP]
> This work runs parallel to the previously broken down work for `Activity` type Records for PR's #3738, #3752

> [!NOTE]
> IAPP Records are created by two network requests with separate payload. One for when showing the Tabular data, and the second for showing the records. This is why there's a distinction When sharing / loading the types of records to display.

- Add a `Back` button to IAPP records because they are read-only
- Create type aliases for `IappTableRow` and `IappRecord`.
- Break down `recordset-layers.ts` further, reducing cognitive complexity.
- Convert IAPP and `WhatsHere[iapp]` actions into ActionCreators in line with recent refactoring
- remove all`actions.ts` actions that were not imported anywhere.

- Add ability to Cache `IAPP` Recordsets
    - Closes #3742
- Add ability to Used Cached IAPP Records in `Whats Here`
    - Closes #3747
- Add ability to Show cached IAPP Records on the map in Offline context
    - Closes #3746
- Set Default Recordset named to an empty string. When recordset is unnamed, show the Default text `New Recordset - [type]`
     - Closes #3733


app/src/state/reducers/map.ts
```diff
@@ -649,6 +648,56 @@ function createMapReducer(configuration: AppConfig): (MapState, AnyAction) => Ma
...
+          }
+          if (draftState.recordTables?.[action.payload.recordSetID]) {
+            draftState.recordTables[action.payload.recordSetID].rows = action.payload.rows;
+          } else {
+            draftState.recordTables[action.payload.recordSetID] = {};
+            draftState.recordTables[action.payload.recordSetID].rows = action.payload.rows;
+          } // set defaults
+          draftState.recordTables[action.payload.recordSetID].loading = false;
+          return;
+        }
```
LocalNewsTV, line 700: Pre-existing code. Just migrated from Switch to if-else block during the actionCreator conversion

meghna0593: I actually like the switch statements, straightforward and maintainable. However looks like we're already moving towards this approach

meghna0593 (approved):
LGTM

PR #3768, opened 2024-12-31
[INV-3757] Clear Records from a Cache

### LINKED ISSUE #3757: `Delete` recordset cache deletes `all` caches
When deleting a recordset cache in web, all records get deleted instead of ones belonging to a cache.

LocalNewsTV:
# Overview

- Implement a Frequency Map on Cache clear actions to avoid deleting records shared with other recordsets.
- Fix issue where clearing `a` cache deletes `ALL` cached records.
- Remove functionality around Cachelists as they were unused.
- Trigger Cache delete when a Recordset is deleted to avoid dead recordsets being stored.
- Closes #3757 

## EDITS
- Convert RecordSet.remove to `Recordset.requestRemoval` now using Thunk.
- Paginate deletion of records in Sqlite in batches of 100 with transactions
    - Tested caching 1000 records, Ran with an arbitrary error being thrown after 300 deletes to test transaction handling.
    - Validated Record counts in CLI

### DIFF
--- a/app/src/UI/Overlay/Records/Records.tsx
+++ b/app/src/UI/Overlay/Records/Records.tsx
@@ -5,12 +5,13 @@ import { OverlayHeader } from '../OverlayHeader';
 import { useHistory } from 'react-router-dom';
 import { useDispatch, useSelector } from 'utils/use_selector';
 import UserSettings from 'state/actions/userSettings/UserSettings';
-import { RecordSetType } from 'interfaces/UserRecordSet';
+import { RecordSetType, UserRecordCacheStatus } from 'interfaces/UserRecordSet';
 import Prompt from 'state/actions/prompts/Prompt';
 import RecordSetDetails from './RecordSetDetails';
 import RecordSetControl from './RecordSetControl';
 import { MOBILE } from 'state/build-time-config';
 import filterRecordsetsByNetworkState from 'utils/filterRecordsetsByNetworkState';
+import RecordCache from 'state/actions/cache/RecordCache';
 
 export const Records = () => {
   const DEFAULT_RECORD_TYPES = ['All InvasivesBC Activities', 'All IAPP Records', 'My Drafts'];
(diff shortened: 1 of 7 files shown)


app/src/UI/Overlay/Records/Records.tsx
```diff
@@ -49,7 +50,10 @@ export const Records = () => {
     e.stopPropagation();
     const callback = (userConfirmation: boolean) => {
       if (userConfirmation) {
-        dispatch(UserSettings.RecordSet.remove(set));
+        if (recordSets[set]?.cacheMetadata?.status === UserRecordCacheStatus.CACHED) {
+          dispatch(RecordCache.deleteCache({ setId: set }));
+        }
+        setTimeout(() => dispatch(UserSettings.RecordSet.remove(set)), 250);
```
plasticviking: Why the delay?

app/src/utils/record-cache/sqlite-cache.ts
```diff
@@ -383,7 +283,22 @@ class SQLiteRecordCacheService extends RecordCacheService {
...
+    }
+    const RecordsToTable = {
+      [RecordSetType.Activity]: 'CACHED_RECORDS',
+      [RecordSetType.IAPP]: 'CACHED_IAPP_RECORDS'
+    };
+    const RECORD_TABLE = RecordsToTable[recordSetType];
+    await this.cacheDB.query(
+      // language=SQLite
+      `DELETE FROM ${RECORD_TABLE}
+       WHERE ID IN (${idsToDelete.map(() => '?').join(', ')})`,
```
plasticviking: Have you tested this with huge lists (10,000+)?

app/src/UI/Overlay/Records/Records.tsx
LocalNewsTV: @plasticviking 
Insurance that the deleteCache process is started (and receives the array of IDs) before the Recordset holding the Ids is deleted.

app/src/utils/record-cache/sqlite-cache.ts
LocalNewsTV: @plasticviking 
Definitely did not. I will adjust it to paginate the queries in smaller batches

PR #3772, opened 2025-01-02
Fix BCGW layer visibility issue on Mobile

### LINKED ISSUE #3632: Mobile: BCGW (BC Geographic Warehouse) layers not visible on map when turned on
Z-index issue? This is while online.

### LINKED ISSUE #3735: Mobile: When the BCGW layers are toggled on, they don't get toggled off on log out
When BCGW layers are toggled on by an authenticated user, the state doesn't get reset on log out. The layers persist on the public map.

Image

meghna0593:
# Overview

This PR includes the following changes:

- Implemented a proxy mechanism in the backend to hit the OWS endpoint for Mobile
- Fix for BCGW layer being visible to unauthenticated/unauthorized user
- Closes [3632](https://github.com/bcgov/invasivesbc/issues/3632) and [3735](https://github.com/bcgov/invasivesbc/issues/3735)

## Type of change

- [ ] Version change

### DIFF
--- a/app/src/UI/LegacyMap/helpers/wms-layers.ts
+++ b/app/src/UI/LegacyMap/helpers/wms-layers.ts
@@ -39,6 +39,7 @@ export const refreshWMSOnToggle = (simplePickerLayers2, map) => {
       const visibility = map.getLayoutProperty(layer.url, 'visibility');
       if (visibility !== 'none' && !layer.toggle) {
         map.setLayoutProperty(layer.url, 'visibility', 'none');
+        map.moveLayer(layer.url, LAYER_Z_MID);
       }
       if (visibility !== 'visible' && layer.toggle) {
         map.setLayoutProperty(layer.url, 'visibility', 'visible');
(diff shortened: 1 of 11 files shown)


LocalNewsTV (commented):
Everything looks good from what I am able to test/see! Would prefer if @plasticviking could run it with a physical device if available.

app/src/UI/LegacyMap/helpers/wms-layers.ts
```diff
@@ -47,3 +51,12 @@ export const refreshWMSOnToggle = (simplePickerLayers2, map) => {
     }
   });
 };
+
+export const hideWMSIfUnauthorized = (simplePickerLayers2, map) => {
+  simplePickerLayers2.map((layer) => {
+    if (map.getLayer(layer.url)) {
+      map.setLayoutProperty(layer.url, 'visibility', 'none');
```
plasticviking: In my testing, when a mobile device is offline, setting layer visibility to `none` does not prevent requests from being sent. I think it is preferable to delete the layer and source and re-add them if possible in response to online/offline events.

PR #3774, opened 2025-01-03
Hide Panel on logout

### LINKED ISSUE #3773: Mobile: On logout, the records panel is visible
Image

meghna0593:
## Type of change

- [ ] Version change

### DIFF
--- a/app/src/UI/Overlay/Overlay.tsx
+++ b/app/src/UI/Overlay/Overlay.tsx
@@ -13,12 +13,10 @@ const Overlay = (props) => {
   const panelOpen = useSelector((state) => state.AppMode.panelOpen);
   const fullScreen = useSelector((state) => state.AppMode.panelFullScreen);
 
-  const { authenticated } = useSelector((state) => state.Auth);
-
   return (
     <div
       id="overlaydiv"
-      className={`map__overlay ${authenticated && panelOpen && !fullScreen ? 'map__overlay--show' : ''} ${
+      className={`map__overlay ${panelOpen && !fullScreen ? 'map__overlay--show' : ''} ${
         panelOpen && fullScreen ? 'map__overlay--show-fullscreen' : ''
       }`}
     >
--- a/app/src/UI/Overlay/Overlay.tsx
+++ b/app/src/UI/Overlay/Overlay.tsx
@@ -13,10 +13,12 @@ const Overlay = (props) => {
   const panelOpen = useSelector((state) => state.AppMode.panelOpen);
   const fullScreen = useSelector((state) => state.AppMode.panelFullScreen);
 
+  const { authenticated } = useSelector((state) => state.Auth);
+
   return (
     <div
       id="overlaydiv"
-      className={`map__overlay ${panelOpen && !fullScreen ? 'map__overlay--show' : ''} ${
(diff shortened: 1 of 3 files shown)


app/src/UI/Overlay/Overlay.tsx
```diff
@@ -13,10 +13,12 @@ const Overlay = (props) => {
   const panelOpen = useSelector((state) => state.AppMode.panelOpen);
   const fullScreen = useSelector((state) => state.AppMode.panelFullScreen);
 
+  const { authenticated } = useSelector((state) => state.Auth);
+
   return (
     <div
       id="overlaydiv"
-      className={`map__overlay ${panelOpen && !fullScreen ? 'map__overlay--show' : ''} ${
+      className={`map__overlay ${authenticated && panelOpen && !fullScreen ? 'map__overlay--show' : ''} ${
```
plasticviking: Won't this also hide the access request page and other no-login-required descendants of `Overlay`?

meghna0593: Didn't test the access request page. The other no-login-required descendants were fullScreen based

LocalNewsTV: Because the auth works differently in Capacitor (doesn't refresh on state change) couldn't we just add a history redirect to push logged out users to the home page, which would remove the records from view and all future descendants. It would match what happens in browser.

meghna0593: This change works on mobile and web. 

@LocalNewsTV Could you elaborate on this `add a history redirect to push logged out users to the home page` ?  The issue isn't about redirecting user to the logged out page/view, it is with having the panel (at the bottom) open and details of the latest activity visible.

LocalNewsTV: The contents of Overlay is contextual to the page that its currently on. 
we use the `useHistory` hook to traverse through the app. 

When we login/logout on `Web` the page refreshes, but because we're using`authBridge` on iOS via capacitor, we don't get that refresh, so the Record page remains.

So in modifying the logout button like So
```ts
const LogoutButton = () => {
  const dispatch = useDispatch();
  const history = useHistory();
  return (
    <MenuItem
      onClick={() => {
        dispatch({ type: AUTH_SIGNOUT_REQUEST });
        history.push('/Landing');
      }}
    >
      <ListItemIcon>
        <LogoutIcon />
      </ListItemIcon>
      Logout
    </MenuItem>
  );
};
```
We change the context of the overlay, and Remove the `/Records` context it had before.

But in proving my concept before typing all of this, I noticed that the Map doesn't reset to the public map state, and we may just want to implement a `window.reload()` so everythings context gets reset, including the auth.

LocalNewsTV: The proof of concept for that also exists in our already existing Refresh button
image

Unless physical devices have a different behaviour than the simulator. You both would know that better than I.

LocalNewsTV (dismissed):
- Confirmed The Record panel is no longer accessible after logging out
- Confirmed the Public map layer appears afterwards

Out of Scope for this ticket
- Noticed Recordset layer markers are still visible on map after logout due to layers not being destroyed. Perhaps the native auth needs to be examined in the future, likely more things under the surface.
    - With the ability to work as an `Offline user`, it doesn't really matter.

PR #3775, opened 2025-01-07
[INV-3771] Speed up maptile caching

### LINKED ISSUE #3771: Speed up Maptile caching.
# Describe the task

use Promise concurrency to speed up the download speed of maptiles from 1 to 5-10 at a time

## Value Add
- Every concurrent tile we download increases our download performance by close to 100%

## Acceptance Criteria

- [ ]  Tiles Download Concurrently
- [ ]  Progress Bars update appropriately

## Additional Context
- Hypothetically should be able to download our tiles 500-1000% faster (5-10 concurrent requests) before taking a performance hit.

LocalNewsTV:
# Overview

- Modify the download method of Map tiles to run a Concurrency limit of 10 instead of 1.
- Modify the Download method to retry once on failure (previously 0)
- Fix HTML Semantic Issue about missing `tr` in `thead`
- add toLocaleString to tIlecounts to better display info (adds commas).
- Increased time between Callback updates 200 -> 500ms
- Closes #3771

Its hard to establish how much the download speed will really improve by for Mobile when using a Simulator. Given the Simulator is constrained by my System the speed is much slower than the below.

Testing in Browser cached 205,587 tiles (2.8GiB)  in 33:18.
Originally ~2300 tiles (36.7MiB) would take 2:34. Scaled up that translates to ~3.6 hours

`(205k / 2,300) * 154sec / 60 / 60`

Testing in a Simulator cached 1,000 tiles in 1:37
Originally 1,000 tiles would take 2:15
I'm not testing for high values because a simulator is more throttled than a physical device so the metrics mean less.
Based on the simulator this completes the same number of tiles in 71.85% of  the time. 

**During Caching I periodically navigated through the app testing for any noticeable slowdown, If you have the Network tab open you'll get some, due to how many Network requests are being logged, but if you don't, There isn't any..

### DIFF
--- a/app/src/UI/Overlay/TileCache/TileCacheListRow.tsx
+++ b/app/src/UI/Overlay/TileCache/TileCacheListRow.tsx
@@ -76,7 +76,7 @@ const TileCacheListRow = ({ metadata, visible }) => {
       </td>
       <td>{metadata.description || metadata.id}</td>
       <td>{metadata.status}</td>
-      <td>{stats?.tileCount}</td>
+      <td>{stats?.tileCount?.toLocaleString('en-US')}</td>
       <td>{stats && convertBytesToReadableString(stats.sizeInBytes)}</td>
       <td>
         <IconButton color={'primary'} onClick={handleEditCacheDescription}>
--- a/app/src/UI/Overlay/TileCache/TileCacheListRow.tsx
+++ b/app/src/UI/Overlay/TileCache/TileCacheListRow.tsx
@@ -76,7 +76,7 @@ const TileCacheListRow = ({ metadata, visible }) => {
       </td>
       <td>{metadata.description || metadata.id}</td>
       <td>{metadata.status}</td>
-      <td>{stats?.tileCount?.toLocaleString('en-US')}</td>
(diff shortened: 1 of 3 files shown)


app/src/UI/Overlay/TileCache/TileCacheListRow.tsx
plasticviking, line 1: better to omit `en-US` argument and let browser choose?

app/src/utils/tile-cache/index.ts
```diff
@@ -107,75 +114,108 @@ abstract class TileCacheService {
...
-                  processedTiles,
-                  totalTiles
-                });
-              }
-            }
+            tileUrls.push({ id: spec.id, url: spec.tileURL(x, y, z), x, y, z });
           }
         }
       }
+      const promises = tileUrls.map((config) => () => this.downloadTile(config));
```
plasticviking, line 174: nice

app/src/UI/Overlay/TileCache/TileCacheListRow.tsx
LocalNewsTV: @plasticviking Removed.

PR #3780, opened 2025-01-08
Allow usage of photo from gallery/library

### LINKED ISSUE #3744: Mobile: Allow usage of a photo from iPad gallery
Currently adding a photo will only allow you to take a new picture, users want to be able to select a picture already taken on the iPad.

meghna0593:
# Overview

- Enable multiple selection of photos from gallery
- Process photos by converting webpath blob to dataurl before 
- Check permission for camera/photo gallery and alert accordingly

## Type of change

- [ ] Version change

### DIFF
--- a/app/src/UI/Overlay/Records/Activity/PhotoContainer.tsx
+++ b/app/src/UI/Overlay/Records/Activity/PhotoContainer.tsx
@@ -98,7 +98,7 @@ const PhotoContainer: React.FC<IPhotoContainerProps> = (props) => {
     }
   };
 
-  const chooseMultiplePhotosFromLibrary = async () => {
+  const choosePhotosFromLibrary = async () => {
     try {
       await checkPermissionsAndDisplayInfo(CameraSource.Photos);
       const multiplePhotos = await Camera.pickImages({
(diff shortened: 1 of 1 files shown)


app/src/UI/Overlay/Records/Activity/PhotoContainer.tsx
```diff
@@ -37,8 +37,45 @@ const PhotoContainer: React.FC<IPhotoContainerProps> = (props) => {
...
+          alert('Camera access is denied. Please enable camera permissions in your device settings to take photos.');
+        }
+      }
+
+      if (photoOption === CameraSource.Photos) {
+        if (permissions.photos === 'denied') {
+          alert(
+            'Photo library access is denied. Please enable photo library permissions in your device settings to choose photos.'
+          );
+        }
```
LocalNewsTV: These nested ifs could be reduced down into single `if` statements. 
we also have in-house Alerts and Prompts that aren't thread blocking like the built-in Browser method

meghna0593: For some cases, I'd say it's valuable to have blocking alerts. Changed the alerts to cater to the in-house ones.
 
As for the if statements, I'll keep them as is -- they're checking different cases. `Camera.checkPermissions()` returns both `camera` and `photos` status, irrespective of what is being called.

LocalNewsTV: Sorry, for clarity I meant this! Since short circuiting would prevent checking the second issue.

```ts
      if (photoOption === CameraSource.Camera && permissions.camera === 'denied') {
          // Do something
       }
      if (photoOption === CameraSource.Photos && permissions.photos === 'denied') {
          // Do something
       }
```
It was just a minor suggestion, So no need to go change it (Sonarcloud doesnt flag it in this case).

for added context:
Agree on blocking alerts have value in some cases. Part of the leverage in the `Prompt` callback was to mimic a similar behaviour that was getting lost when we shifted away from browser based alerts. Though its not a perfect 1:1

meghna0593: Appreciate you clarifying. 

The logic in this method is simple, so combining the conditions won't really impact readability. Also, there’s no anticipated need to extend this method in the future -- will combine the `if` conditions 👍

PR #3782, opened 2025-01-08
[INV-3630] cache download stops in progress, start emancipating from Redux State

### LINKED ISSUE #3630: Mobile: Cache download stops while in progress
Not sure what caused this, but the download stopped itself and the cache in progress can't be removed.

LocalNewsTV:
# Details

This PR Includes the following changes:

- Users can cancel a Record cache download that is in progress.
    - Users are notified with a success alert upon termination 
- Record cache metadata is now persisted in the local DB (Previously just Redux)
- Method names changed to align more consistently with maptile (and future) methods
- If user deletes a cache that is in progress of downloading, the cache deletes
- Fix issue where a Download stops in progress immediately
    - This was caused by there being no filter object, causing the bbox functionality to throw an error and crash. This is now fixed. 
- Migrate business logic in async thunks, to appropriate Database methods, decoupling the two.
- Added stub placement for adding ProgressCallback functionality 

> [!IMPORTANT]
> This PR Starts the process of moving Redux Recordset metadata into DB storage. In the meantime both states (Redux and DB) have to be stored. Next iteration will be removing the redux data, keeping one source of truth. You may find weird consistencies and Error checks while both states are being managed.

 
- Closes #3630

## Testing

Tested the Following

> [!TIP]
> Additional Context: We don't download duplicate records, so when deleting we use a frequency map to ensure we don't delete Records used in multiple places
 
- Caching a recordset
- Caching two recordsets with the same filters then deleting one to make sure the second cache/records remains in tact
- Caching two recordsets with different filters then deleting one making sure there was **no cross-repo delete**
- Deleting a **cached recordset** to check cleanup
- Deleting a recordset **in progress** to check cleanup
- Cancelling a download that is in progress

Testing was done in both Firefox and the Simulator
Sum of records and Cache metadata was manually verified and counted through queries in their respective storages.

### DIFF
--- a/app/src/state/reducers/userSettings.ts
+++ b/app/src/state/reducers/userSettings.ts
@@ -201,9 +201,11 @@ function createUserSettingsReducer(configuration: AppConfig): (UserSettingsState
           status: UserRecordCacheStatus.DOWNLOADING
         };
       } else if (RecordCache.requestCaching.rejected.match(action) || RecordCache.deleteCache.rejected.match(action)) {
-        draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
-          status: UserRecordCacheStatus.ERROR
-        };
+        if (action.error.message !== 'Early Exit') {
+          draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
+            status: UserRecordCacheStatus.ERROR
+          };
+        }
       } else if (RecordCache.requestCaching.fulfilled.match(action)) {
         draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
           status: UserRecordCacheStatus.CACHED,
--- a/app/src/state/reducers/userSettings.ts
+++ b/app/src/state/reducers/userSettings.ts
@@ -201,7 +201,11 @@ function createUserSettingsReducer(configuration: AppConfig): (UserSettingsState
           status: UserRecordCacheStatus.DOWNLOADING
         };
       } else if (RecordCache.requestCaching.rejected.match(action) || RecordCache.deleteCache.rejected.match(action)) {
-        if (action.error.message !== 'Early Exit') {
+        if (action.error.message === 'Early Exit') {
+          draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
+            status: UserRecordCacheStatus.NOT_CACHED
+          };
+        } else {
           draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
             status: UserRecordCacheStatus.ERROR
           };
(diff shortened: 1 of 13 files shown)


plasticviking:
Looks reasonable to me

app/src/state/reducers/userSettings.ts
```diff
@@ -201,18 +201,24 @@ function createUserSettingsReducer(configuration: AppConfig): (UserSettingsState
           status: UserRecordCacheStatus.DOWNLOADING
         };
       } else if (RecordCache.requestCaching.rejected.match(action) || RecordCache.deleteCache.rejected.match(action)) {
-        draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
-          status: UserRecordCacheStatus.ERROR
-        };
+        if (action.error.message === 'Early Exit') {
+          draftState.recordSets[action.meta.arg.setId].cacheMetadata = {
+            status: UserRecordCacheStatus.NOT_CACHED
+          };
```
LocalNewsTV, line 207: Checks like these are for while we are managing the Two states [Redux / DB] And will be removed in next iteration

app/src/utils/record-cache/index.ts
```diff
@@ -54,29 +69,92 @@ abstract class RecordCacheService {
...
+    });
+
+    if (spec.recordSetType === RecordSetType.Activity && (await this.downloadActivity(args))) {
+      Object.assign(responseData, await this.loadRecordsetSourceMetadata(spec.idsToCache));
+    } else if (spec.recordSetType === RecordSetType.IAPP && (await this.downloadIapp(args))) {
+      Object.assign(responseData, await this.loadIappRecordsetSourceMetadata(spec.idsToCache));
+    } else {
+      this.deleteRepository(spec.setId);
+      throw Error('Early Exit');
+    }
```
LocalNewsTV, line 129: > The thrown error is temporary until Second pass is completed

Download functions return a true on success (Early exit is considered a failure)

We intentionally throw an Error in the false since the cache was abandoned and we don't want to update the Redux state. Now the Thunk will fail, and this specific error gets consumed so the user is none the wiser

LocalNewsTV (commented):
Highlighted *some* of the weird code that is there until the second pass is complete.

This is done in two passes to 
- Avoid a massive PR with differing goals
- Keep focus on the original goal here: 
    - Caching to Db 
    - removing a bug in the downloads 
    - being able to stop a download in progress

plasticviking:
Looks good to me

meghna0593:
looks good! Tested on the simulator and ipad

PR #3810, opened 2025-01-27
Add Pause, Resume, Cancel functionality and Progress Bar for IAPP and Activity Download

meghna0593:
# Overview

- Implement a Pause/Resume button to halt and continue the caching process
- Integrate a progress bar to show real-time caching progress
- Include a Cancel button to stop the ongoing caching task

## Type of change

- [ ] Version change

## How Has This Been Tested?

Tested on the simulator and browser (mobile mode)


app/src/UI/Overlay/Records/ProgressControlPanel.tsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
```
LocalNewsTV: Better to destructure what you need instead of importing all of React.
It helps add top level communication about what to expect in a file without needing to dig into it :) 

This is one of my common cleanups when I'm working

app/src/UI/Overlay/Records/RecordSetCacheButtons.tsx
```diff
@@ -16,6 +18,15 @@ const RecordSetCacheButtons = ({ recordSet, setId }: PropTypes) => {
...
   const [cacheActionEnabled, setCacheActionEnabled] = useState<boolean>(false);
 
+  const downloadProgress = useSelector(
+    (state) => state.UserSettings?.recordSets[setId].cacheDownloadProgress,
+    shallowEqual
+  );
+  const activeDownloads = downloadProgress.normalizedProgress != 0;
+
+  // Ensure the pause/resume button reflects the correct state if the user refreshes after pausing
+  const [isPaused, setIsPaused] = useState(activeDownloads || false);
```
LocalNewsTV: activeDownloads is already a pre-declared boolean, shouldn't need to have a fallback on it. Could also have a declared type `useState<boolean>`. 
Would recommend keeping the `useStates` and `useSelectors` grouped together for organization

```ts
const RecordSetCacheButtons = ({ recordSet, setId }: PropTypes) => {
  const dispatch = useDispatch();
  const connected = useSelector((state) => state.Network.connected);
  const downloadProgress = useSelector(
    (state) => state.UserSettings?.recordSets[setId].cacheDownloadProgress,
    shallowEqual
  );
  const activeDownloads = downloadProgress.normalizedProgress !== 0;
  // Ensure the pause/resume button reflects the correct state if the user refreshes after pausing
  const [isPaused, setIsPaused] = useState<boolean>(activeDownloads);
  const [cacheActionEnabled, setCacheActionEnabled] = useState<boolean>(false);
```

app/src/state/actions/cache/RecordCache.ts
```diff
@@ -37,17 +46,30 @@ class RecordCache {
...
 
       return {
         setId: spec.setId,
-        status: downloadCompleted ? UserRecordCacheStatus.CACHED : UserRecordCacheStatus.NOT_CACHED
+        status:
+          downloadMode === CacheDownloadMode.ABORT
+            ? UserRecordCacheStatus.NOT_CACHED
+            : downloadMode === CacheDownloadMode.PAUSE
+              ? UserRecordCacheStatus.PAUSED
+              : UserRecordCacheStatus.CACHED
```
LocalNewsTV: I'd decouple this nested ternary from the return value and declare it above using a `IIFE`

app/src/utils/record-cache/index.ts
```diff
@@ -12,10 +12,18 @@ export enum IappRecordMode {
   Record = 'record',
   Row = 'row'
 }
+
+export enum CacheDownloadMode {
+  DEFAULT = '',
+  PAUSE = 'pause',
+  ABORT = 'abort'
```
LocalNewsTV, line 19:
```suggestion
  DEFAULT,
  PAUSE,
  ABORT
```

As an FYI, not assigning values just auto-assigns them `0, 1, 2...` 
I generally just use strings if I want to use the values for something in the UI.

meghna0593: Good to know! 
I like the one that's in place now, it's more descriptive and understandable.

app/src/UI/Overlay/Records/ProgressControlPanel.tsx
meghna0593: older versions required one to add this import. For React 17 and later, you don't need to import it -- removing it :)

app/src/UI/Overlay/Records/RecordSetCacheButtons.tsx
meghna0593: I think keeping similar logic together makes it readable too

PR #3814, opened 2025-01-28
[INV-3808] Unhandled Promise rejections in Auth

### LINKED ISSUE #3808: Access Request Page API Crash
Some API call made by the Access Request page is crashing the API server (with an unhandled exception).

LocalNewsTV:
# Overview

- reduce all the error code in Auth to be more DRY using `rejectWithErr`
- Modify thrown objects to be errors instead of regular objects
- Handle Rejected Promises (Source of errors)
- Reduce code smells
- Closes #3808

Original User with issue was able to resolve their malformed tokens and they are in working order now, these changes help strengthen our API to handle these issues as they come in, and not break.


plasticviking (approved):
👍🏻

Issue #3739, opened 2024-12-11
Chemical treatments submitted without required information


brennanwebster:
Example activity_incoming_data_id 204186, this is a Draft version of an activity that has the required information in the activity_payload. The next version that was Submitted is 204187, which is missing the calculation_results object in the activity_payload.

This is one instance of something that has been happening fairly frequently, which is Chemical Treatment records somehow being Submitted without required information (activities created through the form are effected by this issue and batch records are not). What information is missing varies but it is usually the herbicides, invasive_plants, or calculation_results objects in the activity_payload. I have contacted the users who have created these records to re-enter the information and they have done so successfully. 

I have not been able to replicate the issue myself, even entering the exact same information as a broken chemical treatment record does not cause the issue to happen. 

The reason I notice this issue when it happens is the missing information in Submitted records causes the database-refresh-views cronjob to fail to refresh the chemical treatment view. Normally I set the broken record to Draft status and email the user to re-enter the information, and then manually refresh the view.

This issue only began happening in the last few months, so I suspect some recent change broke the rube goldberg machine that is the chemical treatment form.

LocalNewsTV:
@brennanwebster This needs testing in Dev before we can deploy

PR #3817, opened 2025-01-29
[INV-3706][INV-3739] Form Rendering Bugfix, Chemical Treatment Fix

### LINKED ISSUE #3706: Rendering issue with form Validation
# Describe the task

There is an issue with component rerendering in the form when the errors change, that cause the form to persist errors that no longer apply, until a re-render is formed for that particular component

## Value Add
- Rendering properly is important for usability of the application, and reduces spaghetti code in place to circumvent the errors

## Acceptance Criteria

- [ ] Forms Rerender appropriately when errors no longer apply

## Additional Context
- The key example of this is the Pest Management Plan fields. Examples and bandaid solution can be found in [PR #3705](https://github.com/bcgov/invasivesbc/pull/3705)

LocalNewsTV:
# Overview

> [!IMPORTANT]
> There is no changes to the methods responsible for Calculations around Chemical treatments. Changes are on the components that use them.

High range of fixes, tune ups and cleaning in Forms related to Chemical Treatments

**Some Issues noted and repaired during work**
- When re-rendering some form components (via hotreload), data would disappear from inputs, suggesting records weren't updating correctly. when they should.
    - JSON.parse(JSON.stringify(form_data)) creating shallow copy
    - Revert Bandaid fix from PR #3705.
- Forms could be submitted without necessary calculations as the error `Couldn't perform calculation because of the invalid scenario.` was non-blocking.
    - Check length of return object from Chemical Calculations, blocking when empty (no calculations). 

**General Fixes**
- Move Chemical treatment Redux action into an `actionCreator` inline with ongoing overhaul efforts
- Clean up CSS so Chemical Treatment Forms aren't squished together
- Clean up Code smells / General Code cleanup:
    - Replacing Nested Ternarys with IIFE functions
    - Replacing `bracket[notation]` with `dot.notation`, where applicable
    - Adding Proper typing to useStates
    - Eliminating `React` imports, using destructuring instead
    - Moving `ErrorListTemplate` to own file
    - Remove unused Variables, replace with `_` when position is needed
    - Remove commented out code
    - separate *some* anonymous functions from JSX into named functions
    - Convert `explicit` returns to `implicit` where possible.
    - Replace generic `props` with `PropTypes` where applicable
    - replace index keys where replacement doesn't require further refactoring.
    - Remove references to `darkTheme` as is tech debt to App v1, and not in present day app.

- Closes #3706 
- Related to #3739, but requires additional testing to see if other loopholes available.


plasticviking (dismissed):
Tests ok for me and seems like a step in the right direction

meghna0593 (approved):
LGTM

PR #3822, opened 2025-02-04
[INV-3778] Apply Concurrency to Record downloads

### LINKED ISSUE #3778: Speed up Record Cache Speed
# Describe the task

Add Concurrency to the RecordCache downloads to improve the download speed

## Value Add
- Faster download speeds

## Acceptance Criteria

- [ ] Can download multiple records concurrently instead of singly

## Additional Context
- Need to make sure we don't bog down our backend resources with the work
- Similar work done for Tile Caching

LocalNewsTV:
# Overview

- Move `processNext` to BaseCacheService as protected member, removing from TileCache
- Add concurrency to Record Downloads 
- Closes #3778


plasticviking (approved):
👍🏻

PR #3826, opened 2025-02-05
Improve Error Handling for Offline Sync Status and DB Connection Errors

### LINKED ISSUE #3743: Mobile: Form data json showing up in the Offline Sync Status window
After creating an activity online and going into the sync status window users will sometimes see what looks like part of the form data json.
Image

meghna0593:
# Overview

This PR includes the following proposed change:

- Added try/catch block to handle database connection errors more effectively.
- The `getDBConnection` method throws an error when the connection fails, therefore the (previous) return block is never reached. Errors are caught in the catch block, preventing a null return. Removed the redundant `(!connection)` checks
- Updated the offline sync status window to show a generic error message instead of exposing raw JSON payloads
- Did some code cleanup

## Type of change

- [ ] Version change


api/src/paths/batch.ts
```diff
@@ -120,28 +112,31 @@ function listBatches(): RequestHandler {
...
+    } catch (error) {
       return res.status(503).json({
         message: 'Database connection unavailable',
         request: req.body,
         namespace: 'batch',
         code: 503
       });
+    } finally {
+      connection?.release();
     }
```
LocalNewsTV: Terminating the connection here will end make it inaccessible on `line 174`

LocalNewsTV (commented):
One small fix needed.
Doesn't appear to be any other issues. Would prefer a second opinion in case I missed something else.

api/src/paths/batch.ts
meghna0593: good catch! Yes, it will be inaccessible

plasticviking:
Looks good to me, let's try and exercise it in dev before we release it

PR #3832, opened 2025-02-10
[INV-3795] Prefilter Records before caching to remove duplicates

### LINKED ISSUE #3795: Pre-filter Cache Record lists to avoid downloading already cached record
# Describe the task
- Prefilter IDs in a recordset requested for caching, against IDs already cached in the database to avoid unnecessary downloads of records we already have stored. 

## Value Add
- Shrinks download time for subsequent recordsets, since it won't request records we already have.

## Acceptance Criteria

- [ ] If two records are identical, caching completes without any downloads
- [ ] If recordsets have overlap, only uncached records are downloaded.

LocalNewsTV:
# Overview

- Rename two Thunk actions to be inline with current naming convention
- Implement functionality to fetch all IDs currently cached, and filter them for inclusive/exclusive matches
- Modify localForage `deleteCachedRecordsFromIds()` to remove ESLint errors 
- Closes #3795

### DIFF
--- a/app/src/utils/record-cache/index.ts
+++ b/app/src/utils/record-cache/index.ts
@@ -1,8 +1,8 @@
+import { GeoJSONSourceSpecification } from 'maplibre-gl';
 import IappRecord from 'interfaces/IappRecord';
 import IappTableRow from 'interfaces/IappTableRecord';
 import UserRecord from 'interfaces/UserRecord';
 import { RecordSetType, UserRecordCacheStatus } from 'interfaces/UserRecordSet';
-import { GeoJSONSourceSpecification } from 'maplibre-gl';
 import { getCurrentJWT } from 'state/sagas/auth/auth';
 import { getSelectColumnsByRecordSetType } from 'state/sagas/map/dataAccess';
 import BaseCacheService from 'utils/base-classes/BaseCacheService';
@@ -111,6 +111,8 @@ abstract class RecordCacheService extends BaseCacheService<
     limit: number
   ): Promise<IappRecord[]>;
 
+  protected abstract getAllCachedIds(): Promise<string[]>;
+
   public abstract isCached(repositoryId: string): Promise<boolean>;
 
   public abstract getIdList(repositoryId: string): Promise<string[]>;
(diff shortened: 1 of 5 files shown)


app/src/utils/record-cache/index.ts
```diff
@@ -188,23 +190,22 @@ abstract class RecordCacheService extends BaseCacheService<
     // IAPP Records use O(n*2) Requests compared to activities.
     const IAPP_CONCURRENCY_LIMIT = Math.ceil(this.CONCURRENCY_LIMIT / 2);
     const executing: Set<Promise<void>> = new Set();
-
+    const uncachedRecords = await this.filterIds('exclusive', spec.idsToCache);
+    const lengthDifference = spec.idsToCache.length - uncachedRecords.length;
     let pauseOrAbort: CacheDownloadMode = CacheDownloadMode.DEFAULT;
-    let processedCaches = spec.processedActivities;
+    let processedCaches = lengthDifference;
```
meghna0593: instead of having two separate variables here, we can just have `let processedCaches = spec.idsToCache.length - uncachedRecords.length;` 
Easy to understand and removes redundancy

app/src/utils/record-cache/index.ts
```diff
@@ -188,23 +190,22 @@ abstract class RecordCacheService extends BaseCacheService<
...
+    const lastProgressCallback: null | number = null;
+    const totalRecordsToCache = uncachedRecords.length;
 
-    let lastProgressCallback: null | number = null;
-    let totalRecordsToCache = spec.idsToCache.length;
```
meghna0593: And we can keep `totalRecordsToCache=spec.idsToCache.length` since we're subtracting on line 194 and adding it on line 251.

meghna0593:
Just a couple of changes to keep it simple. Otherwise, the PR looks good

app/src/utils/record-cache/index.ts
LocalNewsTV: Between Friday and today I changed a lot of the implementation, thanks for catching that looming tech debt! :)

PR #3842, opened 2025-02-12
Display All Offline Activities Simultaneously on the Map

meghna0593:
# Overview

This PR includes the following proposed change:

- Allows users to view all offline activities on the map simultaneously
- The shapes remain visible across all map views, except when new activities are created or when a current activity is requested
- In the next iteration, this feature will be incorporated into the `Records` tab with a more polished user experience

## Type of change

- [ ] Version change


app/src/UI/LegacyMap/helpers/components/CurrentActivityLayer.tsx
```diff
@@ -1,4 +1,5 @@
 import { useContext, useEffect, useState } from 'react';
+import { selectOfflineActivity } from 'state/reducers/offlineActivity';
```
LocalNewsTV, line 2: Unused Import

app/src/state/actions.tsx
```diff
@@ -69,6 +69,7 @@ export const ACTIVITY_RESTORE_OFFLINE = 'ACTIVITY_RESTORE_OFFLINE';
 export const ACTIVITY_OFFLINE_DELETE_ITEM = 'ACTIVITY_OFFLINE_DELETE_ITEM';
 export const ACTIVITY_UPDATE_SYNC_STATE = 'ACTIVITY_UPDATE_SYNC_STATE';
 export const ACTIVITY_OFFLINE_SYNC_DIALOG_SET_STATE = 'ACTIVITY_OFFLINE_SYNC_DIALOG_SET_STATE';
+export const ACTIVITY_OFFLINE_ALL_SHAPE_VISIBILITY_STATE = 'ACTIVITY_OFFLINE_ALL_SHAPE_VISIBILITY_STATE';
```
LocalNewsTV, line 72: We're trying to slowly remove this file and way of handling actions. Further development in it is a step backwards

app/src/state/actions/activity/Offline.ts
LocalNewsTV, line 1: Additional Context on this file, 
It was stubbed out to try and form logical groups of the actions from `actions.tsx`. They're actually unused in the app as they haven't been refactored (yet) and the priority of it was mildly pulled away shortly after the initial work was done.
You can verify this by doing a repo search for 98% of them.  

In general, the class should get a `Prefix` value added, and be implemented in your new action.

We also access this class via `Activity` so `Activity.Offline.setAllShapeVisibility`

app/src/state/reducers/offlineActivity.ts
```diff
@@ -109,6 +112,10 @@ function createOfflineActivityReducer(
             draftState.statusDialogOpen = action.payload.open;
             break;
           }
+          case ACTIVITY_OFFLINE_ALL_SHAPE_VISIBILITY_STATE: {
+            draftState.offlineActivitiesVisibility = action.payload.toggle;
+            break;
+          }
```
LocalNewsTV, line 118: Should follow the action Creator path, instead of the string path.

app/src/UI/OfflineDataSync/OfflineDataSyncTable.tsx
```diff
@@ -114,8 +114,27 @@ export const OfflineDataSyncTable = () => {
...
+                onClick={() => {
+                  dispatch({
+                    type: ACTIVITY_OFFLINE_ALL_SHAPE_VISIBILITY_STATE,
+                    payload: { toggle: !offlineActivitiesVisibility }
+                  });
+                }}
+                color="primary"
+              >
+                {offlineActivitiesVisibility ? <Layers /> : <LayersClear />}
+              </IconButton>
```
LocalNewsTV, line 133: A nameless icon next to submit that only appears when `offline` isn't very intuitive to use. Users on an iPad don't get the benefit of ToolTips that is expected here.

You could make use of the Layer Picker, or creating a component to leverage the existing Records Layer and sit above `Draft`

app/src/state/sagas/activity.ts
```diff
@@ -115,13 +117,23 @@ function* handle_ACTIVITY_SET_CURRENT_HASH_REQUEST(action) {
...
       id = afterColon.includes('/') ? afterColon.split('/')[0] : afterColon;
     }
+
+    if (offlineActivitiesVisibility) {
+      // turn this off when the user creates new records or checks for current activity while offline
+      yield put({
+        type: ACTIVITY_OFFLINE_ALL_SHAPE_VISIBILITY_STATE,
+        payload: { toggle: false }
+      });
+    }
```
LocalNewsTV, line 135: From an end user perspective it just looks like the application has bugs when the layers keep turning off everytime they open a recordset. There's no notification or communication, they just have to keep opening the sync menu and re-enabling the layer. But when they go back to their work, its gone again.

LocalNewsTV (changes requested):
There's still a few rounds of iteration that should be done here. 
I would recommend a walk through of this with one of the stakeholders. 

If you need additional help with the Redux stuff, I am happy to help.

meghna0593:
> There's still a few rounds of iteration that should be done here. I would recommend a walk through of this with one of the stakeholders.
> 
> If you need additional help with the Redux stuff, I am happy to help.

Adding this to the Layer Picker is a good suggestion. I'll make that change and update the PR. 
All good with redux stuff. The string path approach was taken to maintain consistency for that file.

PR #3850, opened 2025-02-18
[N/A] Raise centerpoint of markers when using quick pan

LocalNewsTV:
# Overview

When using the Quick Pan tool, raise the Latitude of where the map centers.
Presently the point ends up in the center, and is typically covered by the Overlay


plasticviking (approved):
👍🏻

PR #3851, opened 2025-02-18
[INV-3734] Filtering Overhaul (bug fixes, UI changes)

### LINKED ISSUE #3719: Style Filter Inputs in /Record
# Describe the task

Currently they're unstyled, so they get overridden by the iOS Settings, which is smaller and very blue. By styling them we can have a more consistent look across versions of our application, and provide better accessibility of our app by increasing the touch targets.

## Value Add

- Consistent Look and feel
- Easier usability in Mobile

## Acceptance Criteria

- [ ] Filter buttons styled
- [ ] Styes are consistent between what's displayed in Safari/iOS/Firefox/Chrome

## Additional Context

-

### LINKED ISSUE #3734: Mobile: Adding filter to recordset sometimes makes all records disappear when it shouldn't
Sometimes when adding multiple filters to a recordset, a new filter will be added and all the records will disappear before the user types anything.

LocalNewsTV:
# Overview

> [!IMPORTANT]
> Several large diffs are caused by pre-lint hooks in some files, and removal of React.Fragments in some components where Overlay headers were

### CSS/UI Work

 - Remove all instances of `OverlayHeader`, with the exception of `OverlayContent.tsx`
     - This helps the header stay fixed in place,  removing instances where scrolling far enough disappears it.
- Set the height of the Record table footer as variable in App.css. 
    - Helps have dynamic sizing, previously this was getting covered by the iPad Dock bar
- Expand size of Filters, and add styling for consistency between browsers/devices
    - Closes #3719 
- Sticky Headers for tables
- Contain Tables in containers for overflow scrolling. 

### Bug Fixing / Changes

- Modify `Filters.tsx` to take a filter as a prop, maintaining consistency and removing `useSelector` hooks
- Add conditional render to Record table for when no results are returned (previously shows only table headers)
- Adjust Creation of Filters to use Nanoid for id key, instead of relying on array indices.
    - Fixes issues with deleting/updating filters
- Fix issue where debounce function will not fire when value is same as initial value
    - e.g. Starting a new filter, putting in a value triggers a rerender, emptying the value will not rerender
- Add stronger typing around Filters
- Destructure Props in `Recordset.tsx`
- convert filter actions to action creators
- remove in house `getUuid` method, replacing with `nanoid`
- Drastically shrink `UserSettings.RecordSet.updateFilter` reducer to something more succinct
- Closes #3734 

### No Results

image

image

### Filters 

image

image


plasticviking (approved):
looks and tests good to me

PR #3853, opened 2025-02-19
Fix pagination and update created_by for Records

### LINKED ISSUE #3843: New records being created and saved do not show up under Drafts
The `created_by` value is not being populated upon saving, causing the records to not appear on drafts.

meghna0593:
- Fix missing `created_by` value on activity creation 
- Added `SAVE_SUCCESSFUL_PRIVATE` status to monitor records saved to db. Also used to fire up POST/PUT methods
- Fix pagination on Recordset


plasticviking (dismissed):
This will clobber `created_by` on any record modified by the logged-in user (for example, if an admin is editing someone else's record, `created_by` will be overwritten on save). It might make sense to update the SQL query to set created_by = updated_by if and only if created_by is null/empty.

meghna0593:
This PR fixes the issue on [this](https://github.com/bcgov/invasivesbc/issues/3843) ticket. Done some thorough testing.

However, I want to do more clean up work on the workflow we have in place. Created another [ticket](https://github.com/bcgov/invasivesbc/issues/3866) to address that.

app/src/UI/Overlay/Records/RecordSet/RecordSetFooter.tsx
```diff
@@ -9,7 +9,7 @@ const RecordSetFooter = (props) => {
 
   const loading = layer?.loading || recordTable?.loading;
 
-  const totalRecords = layer?.IDList?.length;
+  const totalRecords = recordTable?.rows?.length;
```
LocalNewsTV: We don't fetch all rows at once, we paginate them and call for more with each page turn. With this it will always be 20 records and 1 page. 
image
For reference this recordset (default activity) has 54k records. 

the IDList being replaced was fetched at app load/filter changes, and was a complete ID list for the recordset `string[]`

Expectation:
image

app/src/state/sagas/activity/offline.ts
```diff
@@ -109,18 +110,27 @@ export function* handle_ACTIVITY_RUN_OFFLINE_SYNC() {
...
+        hydrated.sync_status === ActivitySyncStatus.SAVE_SUCCESSFUL_PRIVATE
+          ? yield InvasivesAPI_Call('PUT', `/api/activity/`, {
+              ...hydrated,
+              form_status: ActivityStatus.DRAFT
+            })
+          : yield InvasivesAPI_Call('POST', `/api/activity/`, {
+              ...hydrated,
+              form_status: ActivityStatus.DRAFT,
+              sync_status: ActivitySyncStatus.SAVE_SUCCESSFUL_PRIVATE // saved to db but only visible to user
+            });
```
LocalNewsTV, line 126: ~Since the end result in either scenario is that `sync_status = Save_Successful_private` the `else` condition could just be the only condition.
Unless this was unintended.~
Completely overlooked the Method changing

LocalNewsTV (commented):
~Making a record I was still not getting `created_by` fields in my DB entries. They do show up in the API requests, making it a backend issue.~ Local issue

I also noticed that in Mobile my records despite being submitted (not drafted) were still showing up in the DB as `Draft`, making them unviewable to anyone else. (I can do this in dev though, it is not caused by this branch)

I'm not quite sure what the purpose of `SAVE_SUCCESSFUL_PRIVATE` is. Draft records are already viewable only to the user that is under `created_by`, where Submitted records are viewable by people with sufficient access privilege. Other than being assigned it doesn't seem deterministic on anything.

meghna0593:
> Making a record I was still not getting `created_by` fields in my DB entries. They do show up in the API requests, making it a backend issue.
> 
> I also noticed that in Mobile my records despite being submitted (not drafted) were still showing up in the DB as `Draft`, making them unviewable to anyone else. (I can do this in dev though, it is not caused by this branch)
> 
> I'm not quite sure what the purpose of `SAVE_SUCCESSFUL_PRIVATE` is. Draft records are already viewable only to the user that is under `created_by`, where Submitted records are viewable by people with sufficient access privilege. Other than being assigned it doesn't seem deterministic on anything.

meghna0593:
> Making a record I was still not getting `created_by` fields in my DB entries. They do show up in the API requests, making it a backend issue.
> 
> I also noticed that in Mobile my records despite being submitted (not drafted) were still showing up in the DB as `Draft`, making them unviewable to anyone else. (I can do this in dev though, it is not caused by this branch)
> 
> I'm not quite sure what the purpose of `SAVE_SUCCESSFUL_PRIVATE` is. Draft records are already viewable only to the user that is under `created_by`, where Submitted records are viewable by people with sufficient access privilege. Other than being assigned it doesn't seem deterministic on anything.

that's weird, `created_by` gets populated in my db entries. I haven't tested out `submitted` portion for this PR, the ticket was about Drafts. 
Added `SAVE_SUCCESSFUL_PRIVATE` to have a status for the Activity itself. When Drafts are getting created/edited new versions of them are being generated and their sync_state is `LOCALLY_MODIFIED` always. To keep track of new entries that are not already on db, `SAVE_SUCCESSFUL_PRIVATE` was added.  

As mentioned above, this PR is a quick fix. We need to do rework for the way new records are being created and edited in mobile

app/src/state/sagas/activity/offline.ts
meghna0593: This was intentional. The first time user saves their entry when online, or runs sync, we want to hit the `POST` method. After that, `PUT` is called.
The workflow for mobile, whether online or offline, follows the same methods in offline.ts, which has led to confusion and raised concerns about rework for better maintainability.

app/src/UI/Overlay/Records/RecordSet/RecordSetFooter.tsx
meghna0593: hmm.. I seem to have overlooked the other recordsets for this. I was going over `My Drafts` only. 
The pagination on `My Drafts` shows `No records found` always with the existing code -- updates on refresh only. Will look into this.

LocalNewsTV (approved):

Issue #3796, opened 2025-01-13
[Backend] Update Cached Records functionality


LocalNewsTV:
# Describe the task
> If the task will only make sense to a dev, add a :zap: to the title.

Receive an API request containing a list of IDS and the date a Cache was made. 
have the API Send back a filtered list of Ids where a record was updated AFTER the supplied date

## Value Add
- Recordsets can be updated in curated batches, avoiding complete redownloads of a recordset

## Acceptance Criteria

- [ ] Given I submit an IDList and Date, the API returns the IDList filtered with records AFTER the given date

PR #3890, opened 2025-03-05
[INV-3797] Frontend** Update Activity Records Caches

### LINKED ISSUE #3797: [Frontend] Update Cached records functionality
# Describe the task

Add functionality to update Cached records stored on a device

## Value Add
- Recordsets can be updated in curated batches, avoiding complete redownloads of a recordset

## Acceptance Criteria

- [ ] Given I submit an IDList and Date, the API returns the IDList filtered with records AFTER the given date
- [ ] Given a filtered IDList, download all records contained on the list.
- [ ] When the IDList has been downloaded, update the cacheTime of recordsets to reflect the current date

LocalNewsTV:
# Overview

Complementary to #3796

## Type of change

Making Caches updates daily for users to avoid deleting/recaching offline recordsets.

- Once per day, when the user authenticates, update the record caches and Repository 
    - Time of Cache
    - Cached GeoJSON/Centroids
    - idLists of recordsets
- Add Alerts to communicate success/failure of Updating Record caches
- Repository metadata now stores the filterObjects from a Recordset
    - This lets us check for New records we don't already have cached, in addition to any that have changed 
- Add new Methods:
   - isSameDay **(util)**
   - getListOfNewIds **(recordCaching)**
   - updateActivityCaches **(recordCaching)**
- Add Interface for FilterObjects
- move BATCH_AMOUNT to class member instead of scoped to functions (reduces Activity pulls by 2)
- Closes #3797

### DIFF
--- a/app/src/utils/record-cache/sqlite-cache.ts
+++ b/app/src/utils/record-cache/sqlite-cache.ts
@@ -14,53 +14,11 @@ import {
   CacheDownloadMode
 } from 'utils/record-cache/index';
 import { sqlite } from 'utils/sharedSQLiteInstance';
+import MIGRATIONS from './migrations';
 
 const CACHE_DB_NAME = 'record_cache.db';
 const CACHE_UNAVAILABLE = 'cache not available';
-//language=SQLite
-const RECORD_CACHE_DB_MIGRATIONS_1 = [
-  `CREATE TABLE CACHE_METADATA
-   (
-     SET_ID VARCHAR(64) NOT NULL UNIQUE PRIMARY KEY
-   );`,
-  `CREATE TABLE CACHED_RECORDS
-   (
-     ID   VARCHAR(64) NOT NULL UNIQUE PRIMARY KEY,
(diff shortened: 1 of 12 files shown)


plasticviking:
It works, but it might be more complex than required and therefore potentially fragile. Unless there's something I'm missing about this, why not just use HTTP's cache expiration semantics (eg `Last-Modified` and `If-Modified-Since`)? Then the browser can handle most of this logic for you.

app/src/utils/record-cache/sqlite-cache.ts
```diff
@@ -355,6 +313,7 @@ class SQLiteRecordCacheService extends RecordCacheService {
     const short_id = (data as Record<PropertyKey, Feature[]>)?.short_id;
     const geometry = (data as Record<PropertyKey, Feature[]>)?.geometry;
     const map_symbol = (data as Record<PropertyKey, Feature[]>)?.map_symbol;
+    const activityDate = (data as Record<PropertyKey, any>)?.date_created;
```
plasticviking, line 316: what about updates?

PR #3917, opened 2025-03-18
[INV-3918][INV-3919] Silence failed calls offline + remove refresh button 

### LINKED ISSUE #3918: Silence Failed API Calls
Don't post alert when API Call fails if user is an offline user.

### LINKED ISSUE #3919: Remove Refresh Button
Remove the Refresh Page button from App
was originally created as a quick fix for the draw tools bugs, but didn't work on physical iOS devices.

Bug is long fixed, so button no longer has its purpose

LocalNewsTV:
# Overview

- Remove `Hold to Refresh Page` button from Application
    - Was added as a temporary measure while the Draw tool issues were occurring. This is no longer the case. 
    - `useLongPress` hook worked in the Simulator, but doesn't work on physical device as expected.
        > Simply clicking to refresh was not wanted as to avoid accidental clicks from users.

Closes #3919

- Silence Failed API Calls when not authenticated
    - Currently displays an alert if something fails in a fetch, such as a request for treatment ids 
    - When a user is set as `offlineUser` (not authenticated) its expected, and they don't need misc alerts
    - added `console.error` in replacement, since this may fall under dev debugging.

Closes #3918


app/src/UI/Header/RefreshButton.tsx
```diff
@@ -1,17 +0,0 @@
-import { Refresh } from '@mui/icons-material';
-import { ListItemIcon, MenuItem } from '@mui/material';
-import useLongPress from 'utils/useLongPress';
-const RefreshButton = () => {
-  const handleRefresh = () => location.reload();
```
meghna0593, line 5: ~~Are we sure we want to completely get rid of this? We can add `window.location.reload()` to make it work on ios.~~

confirmed with @brennanwebster - he's fine with removing this button.

meghna0593 (approved):

PR #3931, opened 2025-03-19
[INV-3921][INV-3926] DrawTools** Make inactive shapes Solid line, Get Delete button working

### LINKED ISSUE #3921: Mobile: Delete Button on Draw Tools Not Functioning
When a shape is drawn on the map and the delete option is used, the shape is not deleted. However, when new shapes are drawn, the new ones correctly override the old shape, which is the expected behaviour.

Decision to be made whether to fix the delete button or remove it entirely.

### LINKED ISSUE #3926: Mobile: Dotted line persists after drawing a shape for Tile Caching
Dotted line should change to a solid line once the drawing is completed. 

Image

LocalNewsTV:
# Overview

- When shape is inactive, make solid line instead of dotted
    - Color changed to [BC Gold Color](https://www2.gov.bc.ca/gov/content/digital/design-system/foundations/colour)
        - More neutral than red, and still visible on all maps 
- Simplify the delete button to clear all shape off the map (we only keep one at a time)
    - Delete Shape from Activity (Prompt for confirmation)
- Change the strings in some MapTile actions to match their function name. (Making cross referencing easier and faster)

- Closes #3921
- Closes #3926

### DIFF
--- a/app/src/UI/LegacyMap/helpers/components/DrawControls.tsx
+++ b/app/src/UI/LegacyMap/helpers/components/DrawControls.tsx
@@ -87,6 +87,8 @@ const DrawControls = () => {
     } else {
       if (mode === TargetMode.TILE_CACHE) {
         dispatch(TileCache.clearTileCacheShape());
+      } else if (mode === TargetMode.WHATS_HERE) {
+        dispatch(WhatsHere.clear_whats_here());
       }
       drawInstance.current.deleteAll();
     }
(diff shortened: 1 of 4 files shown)


meghna0593 (commented):
Loving the BC Gold color!
Great enhancements to both the delete tool and Tile Cache. 

Wondering if the delete functionality works with the 'What's Here' feature? For me, the gold color disappears, but the black border remains.

meghna0593:
looks good!

PR #3932, opened 2025-03-19
[INV-3928] Fix time interval between heartbeats

### LINKED ISSUE #3928: Mobile: When out of service, cached recordsets won't display unless app is set to offline mode
This stumped me for a minute when I was field testing so I guarantee it will stump some users. Should there be a message pop-up telling the user to toggle to offline to display cached records on map? Discuss.

LocalNewsTV:
# Overview

- Fix issue where time was not being reduced properly between failed heartbeats causing each check to last a full minute or longer,
- Heartbeats now go 60 seconds per beat, or 20 seconds per beat if there is a currently failing beat on the number (original intention)

- If attempting to fetch table rows from a recordset but the user has spotty internet, attempt to grab records from the cache (if exists).

Possible Fix for #3928, pending additional details

### DIFF
--- a/app/src/state/sagas/map/dataAccess.ts
+++ b/app/src/state/sagas/map/dataAccess.ts
@@ -25,7 +25,7 @@ import { selectUserSettings } from 'state/reducers/userSettings';
 import getSelectColumnsByRecordSetType from 'sharedAPI/src/getSelectColumnsByRecordSetType';
 import { PayloadAction } from '@reduxjs/toolkit';
 import IappActions, { IappTableRowRequest } from 'state/actions/activity/Iapp';
-import Activity from 'state/actions/activity/Activity';
+import Activity, { ActivityTableRowGetRequest } from 'state/actions/activity/Activity';
 
 export function* handle_ACTIVITIES_GEOJSON_GET_REQUEST(action) {
   try {
@@ -224,7 +224,7 @@ export const getRecordFilterObjectFromStateForAPI = (recordSetID, recordSetsStat
   } as any;
 };
 
-export function* handle_ACTIVITIES_TABLE_GET_ROWS(action) {
+export function* handle_ACTIVITIES_TABLE_GET_ROWS(action: PayloadAction<ActivityTableRowGetRequest>) {
   try {
     const currentState = yield select(selectUserSettings);
     const connected = yield select(selectNetworkConnected);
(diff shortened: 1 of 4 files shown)


app/src/state/sagas/map/dataAccess.ts
```diff
@@ -261,25 +261,34 @@ export function* handle_ACTIVITIES_TABLE_GET_ROWS(action) {
...
+      Activity.getRowsSuccess({
+        recordSetID: recordSetID,
+        rows: records,
+        tableFiltersHash: tableFiltersHash,
+        page: page,
+        limit: limit
+      })
+    );
+  } catch {
+    // do nothing. An expected error here is 'repository n not found' if the recordset contains no cache.
```
meghna0593: Consider adding a log statement here instead of leaving the catch block empty. Using `console.error(e)` would help capture unexpected errors and make debugging easier.

LocalNewsTV: @meghna0593  Fair, better to have and not need than need and not have.

meghna0593: ^ a quote worthy of being framed

meghna0593:
Works well. Gets me back online when network is available. 

We should probably dispatch `AuthActions.makeOfflineUserCurrent` too?

app/src/state/sagas/network.ts
```diff
@@ -31,10 +31,9 @@ const canConnectToNetwork = async (url: string): Promise<boolean> => {
 const getTimeBetweenTicks = (heartbeatFailures: number): number => {
   const BASE_SECONDS_BETWEEN_FAILURE = 20;
   const BASE_SECONDS_BETWEEN_SUCCESS = 60;
-  return Math.max(
-    BASE_SECONDS_BETWEEN_FAILURE * 1000 * Math.pow(1.1, heartbeatFailures),
-    BASE_SECONDS_BETWEEN_SUCCESS * 1000
-  );
+  return heartbeatFailures === 0
```
plasticviking: This seems harder to intuitively understand than the previous implementation?

LocalNewsTV:
@meghna0593 
There is no need. The exact same information was parsed from the users token when they logged in, making the `makeOfflineUserCurrent` call wouldn't change anything.

Please review the reducer case for `makeOfflineUserCurrent` and the function `loadCurrentStateFromIdToken`

PR #3934, opened 2025-03-19
Prevent spatial autofill on intersection & fix Line tool behaviour during record creation

### LINKED ISSUE #3784: Mobile: Self-intersecting polygon validation warns the user but doesn't prevent submission
This actually happens on web also. I believe it used to work differently, where if the polygon didn't pass the validation the spatial autofill fields would not fill, so the user would have to try to draw the shape again.

As it is now a warning message pops up in the top right of the map, but the spatial autofill fields will still fill, and then the user will be able to submit the record. But once they are back online and try to sync the record will fail to sync because of the bad geometry. 

I think we need to go back to the old way of forcing the user to create a valid shape by not filling the spatial autofill fields until the shape is valid.

### LINKED ISSUE #3922: Mobile: The line tool exhibits inconsistent behavior when creating a new record
When the user selects the line tool, the area is populated automatically (as expected). However, clicking outside the line causes the area value to disappear.

If the user does not click outside the line, the line disappears upon saving the form. If the user selects the record from Drafts/Unsynced Records and zooms in on the area, the line reappears.

This issue needs to be resolved for consistent functionality.

Image
Image

meghna0593:
# Overview

- Prevents spatial autofill for LineString and Polygon when shape intersections occur during new record creation. Fixes #3784.
- Fixes Line tool behaviour during record creation: clicking outside no longer updates the area or shape. Fixes #3922 

## Type of change

- [ ] Version change

### DIFF
--- a/app/src/state/sagas/activity/dataAccess.ts
+++ b/app/src/state/sagas/activity/dataAccess.ts
@@ -169,6 +169,7 @@ export function* handle_ACTIVITY_UPDATE_GEO_REQUEST(action: Record<string, any>)
       const hasSelfIntersections = kinks(sanitizedGeo.geometry).features.length > 0;
       if (hasSelfIntersections) {
         yield put(Alerts.create(mappingAlertMessages.containsIntersections));
+        return;
       }
     }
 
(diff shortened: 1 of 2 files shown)


LocalNewsTV (approved):
Works as it was meant to :)

PR #3962, opened 2025-03-25
[INV-3948] Add cluster layer to map

### LINKED ISSUE #3948: Add Clustering to Offline Map Layers
# Describe the task

Add Clustering to the Offline Record Set layers 

## Value Add
- Reduced Memory consumption at higher levels from trying to load `n` shapes

## Acceptance Criteria

- [ ] Shapes Cluster at higher level
- [ ] Cluster bubbles appropriately sized
- [ ] Clusters can change colors along with the recordsets
- [ ] Layers created for clustering the cachedCentroid layers
- [ ] Works on IAPP and Activities

## Additional Context

-

LocalNewsTV:
# Overview

- turn the Offline CachedCentroid layer into a cluster display to help reduce panning load
    - Sadly does not reduce weight of initial load 

Discussed and Demod PoC with @brennanwebster 
- Closes #3948

image

### DIFF
--- a/app/src/UI/LegacyMap/helpers/functional/recordset-layers.ts
+++ b/app/src/UI/LegacyMap/helpers/functional/recordset-layers.ts
@@ -156,6 +156,18 @@ const getCircleMarkerZoomedOutLayer = (layerID: string, options: LayerOptions): 
   minzoom: options.minzoom ?? 0
 });
 
+const getClusterLayer = (layerID: string, color): CircleLayerSpecification => ({
+  id: 'cluster-circle-' + layerID,
+  type: 'circle',
+  source: layerID,
+  filter: ['has', 'point_count'],
+  paint: {
+    'circle-color': color,
+    // 10px < 15 point_count < 15px < 30 point_count < 25px
+    'circle-radius': ['step', ['get', 'point_count'], 10, 15, 15, 30, 25],
+    'circle-opacity': 0.7
+  }
+});
 const getLabelLayer = (layerID: string, options: LayerOptions): SymbolLayerSpecification => ({
   id: 'label-' + layerID,
   source: layerID,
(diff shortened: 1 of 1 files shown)


meghna0593:
Was this demo'ed (apart from PoC) with @brennanwebster ? I like what we have now. With this implementation, the cluster tends to disappear (or move around) when I zoom in, requiring me to pan around to locate the actual point. However, with the current approach, the points remain fixed and don’t shift.

meghna0593 (changes requested):
Was this demo'ed (apart from PoC) with @brennanwebster  ? I like what we have now. With this implementation, the cluster tends to disappear (or move around) when I zoom in, requiring me to pan around to locate the actual point. However, with the current approach, the points remain fixed and don’t shift.

And I remember this was a concern when I first started - with points shifting as we zoomed in.

LocalNewsTV:
@meghna0593 

Visually there is no difference in the PoC to end implementation. It just needed colour changing to work correctly and some code cleanup.   

> And I remember this was a concern when I first started - with points shifting as we zoomed in.

The only clustering known to me that we have previous to this PR is in the IAPP Record layer on the `public map`. So I am not sure what you're referencing with this comment.

GeoJSON is expensive to pan around on the map, so clustering was a concession to help reduce resource consumption for mobile users. This only applies when a `user is offline`, the online map is not changed from its already established implementation, because the api is supplying vectors tiles matching the filters.

We can hold off on merging this to discuss with @plasticviking about supplying packages of vector tiles. That PR can be used as a soft warmup to our improved caching of Maptiles we discussed on Monday. Being able to efficiently cache the vector tiles would easily replace this and benefit the application in the long run. In that case, I will move it to Draft until then.

<sub> Also realizing I still need to add layers for IAPP </sub>

meghna0593:
I wasn’t referring to clustering itself, but rather an issue with zooming in and the points moving around. I believe it was a bug I initially worked on for public maps (can't recall). Nonetheless, I was talking about how the clusters disappear when you zoom in.

PR #3963, opened 2025-03-25
[INV-3957] Backup cached recordsets on Redux cache clear

### LINKED ISSUE #3957: Populate cached recordsets on reset of redux state
# Describe the task

if a redux cache is deleted, the recordset gets lost so the cache records are no longer accessible. The cached records still exist on the device, and we have a copy of some details for the cache. 

## Value Add
- More robust offline in case of failure

## Acceptance Criteria

- [ ] If the redux state is cleared, cached recordsets get repopulated

## Additional Context

-

LocalNewsTV:
# Overview

>[!NOTE]
> This is is a Failover function, and is not fired unless the redux state was compromised. This allows an the cached records to be accessible if the redux state fails, is cleared, or a migration key breaks the state.

- Add workflow for if a redux cache is cleared, the SQL/localforage db is checked for existing recordsets and recreated
- Added the colorScheme type to the UserRecordSet interface
- Cache the name of a recordset at the time of creating the recordset
- Closes #3957

To manually trigger this, do the following:

1. Name and Make two recordsets, cache only one of them

image

2. Delete your Redux data, seen here in the Developer tools

image

3. Refresh the page. The cached record was backed up

image

### DIFF
--- a/app/src/interfaces/UserRecordSet.ts
+++ b/app/src/interfaces/UserRecordSet.ts
@@ -27,7 +27,21 @@ export enum UserRecordCacheStatus {
 export interface UserRecordSet {
   tableFilters: IFilter[];
   id?: string;
-  color: string;
+  colorScheme?: {
+    Activity_Biocontrol_Collection: string;
+    Activity_Biocontrol_Release: string;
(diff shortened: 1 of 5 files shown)


meghna0593 (approved):
Tested locally, works as expected.

PR #3985, opened 2025-04-01
Hide Offline activities layer and custom layers on logout

### LINKED ISSUE #3943: Custom layers / offline user record layer appear on Public Map
# Describe the task

When not logged in, Toggled on custom layers and offline activity layers are visible on the public map.

## Value Add
- Consistency, security

## Acceptance Criteria

- [ ] Offline Activities not visible on public layer
- [ ] Custom Layers not visible on public layer

## Additional Context

- These shapes are only stored locally, so this is a per-device issue instead of a greater issue

Image

meghna0593:
# Overview

- Hide authenticated layers when logged out. 
- Closes #3943 

## Type of change

- [ ] Version change

### DIFF
--- a/app/src/UI/LegacyMap/helpers/functional/utility-functions.ts
+++ b/app/src/UI/LegacyMap/helpers/functional/utility-functions.ts
@@ -24,12 +24,3 @@ export function safelySetPaintProperty(map: maplibregl.Map, mapLayer: string, pr
     console.error(e);
   }
 }
-
-export const removeLayerIfUnauthorized = (curr_layer, map) => {
-  curr_layer.map((layer) => {
-    if (map.getLayer(layer.url)) {
-      map.removeLayer(layer.url);
-      map.removeSource(layer.url);
-    }
-  });
-};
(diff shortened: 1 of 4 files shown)


app/src/UI/LegacyMap/helpers/functional/utility-functions.ts
```diff
@@ -24,3 +24,12 @@ export function safelySetPaintProperty(map: maplibregl.Map, mapLayer: string, pr
     console.error(e);
   }
 }
+
+export const removeLayerIfUnauthorized = (curr_layer, map) => {
+  curr_layer.map((layer) => {
+    if (map.getLayer(layer.url)) {
+      map.removeLayer(layer.url);
+      map.removeSource(layer.url);
+    }
```
LocalNewsTV: When the arg comes from `[client_boundaries]`, the value `layer.url` is undefined.

the ID for clientBoundaries is defined in `addClientBoundariesIfNotExists` located in `app/src/UI/LegacyMap/helpers/functional/client-boundaries.ts`

It might be better to just have a specific companion function in that file for logical grouping, and keep the original `hideWMSIfUnauthorized` function.

LocalNewsTV (commented):
Offline Activities disappear upon logout as expected,
But the custom layers don't disappear on logout from iOS (They do on web due to the refresh)

meghna0593:
> Offline Activities disappear upon logout as expected, But the custom layers don't disappear on logout from iOS (They do on web due to the refresh)

LocalNewsTV (approved):
🚀

PR #3996, opened 2025-04-03
[3993] Fix Label Toggle on Map (Online Layers)

### LINKED ISSUE #3993: Recordsets enable/label toggle strange behaviour
If you enable a recordset it will display on the map, if you enable the labels they will show on the map. If you disable the layer but not the labels everything disappears (this is good). But then if you enable the recordset again the layer will display on the map but not the labels, even though the label toggle is still enabled. I think if you disable the recordset while the label toggle is enabled that the label toggle should also be turned off. Message me if this doesn't make sense.

LocalNewsTV:
# Overview

- Disable the Label button when Map Layer is turned off.
- Name boolean variables in recordset-layers.ts to be more obvious
    - Modify the label condition to check that the mapLayer is enabled as well
- Closes #3993

## Screenshots

image

### DIFF
--- a/app/src/UI/LegacyMap/helpers/functional/recordset-layers.ts
+++ b/app/src/UI/LegacyMap/helpers/functional/recordset-layers.ts
@@ -583,12 +583,14 @@ export const refreshVisibilityOnToggleUpdate = (storeLayers, map: maplibregl.Map
     });
     matchingLabelLayers?.map((mapLayer) => {
       const visibility = map.getLayoutProperty(mapLayer, 'visibility');
-      const disableLabelLayer = (visibility !== 'none' && !layer.layerState.labelToggle) || !layer.layerState.mapToggle;
-      const enableLabelLayer = visibility !== 'visible' && layer.layerState.labelToggle && layer.layerState.mapToggle;
+      const shouldHideLabelLayer =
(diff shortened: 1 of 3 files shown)


app/src/UI/LegacyMap/helpers/functional/recordset-layers.ts
```diff
@@ -583,11 +583,13 @@ export const refreshVisibilityOnToggleUpdate = (storeLayers, map: maplibregl.Map
     });
     matchingLabelLayers?.map((mapLayer) => {
       const visibility = map.getLayoutProperty(mapLayer, 'visibility');
-      if (visibility !== 'none' && !layer.layerState.labelToggle) {
-        map.setLayoutProperty(mapLayer, 'visibility', 'none');
-      }
-      if (visibility !== 'visible' && layer.layerState.labelToggle) {
+      const disableLabelLayer = (visibility !== 'none' && !layer.layerState.labelToggle) || !layer.layerState.mapToggle;
```
meghna0593: `enableLabelLayer` and `disableLabelLayer` sound like commands or functions rather than boolean conditions.  For the sake of making it more obvious, we can have `shouldShowLabelLayer` and `shouldHideLabelLayer`

meghna0593 (dismissed):

PR #4099, opened 2025-05-20
[INV-4098] Make loggedInOrWorkingOffline based on account_status

### LINKED ISSUE #4088: Site Lists
# Describe the task

Site Lists are sites that are required to be visited in a year, they're a generated report of different sites in the app (IAPP and IBC). In simplest format, they are a recordset containing both types of Records

## User Story
*As a* **Land Manager** *I want to* **create SiteLists in Invasive BC** *so that* **I can share the list with a hired contractor**

## Value Add
- Having SiteLists accessible via app adds user value and takes needs from context switching applications
- Site lists become updatable

## First task

- Let Users simply upload their sitemap. *Converting to Recordset(s) seems simplest, since we have Ids and functionality*

First task workflow:
- Users upload their site list - fit a pre-defined schema 
- System highlights all the records for that user
- Save it as a new layer
-  First step -> uploading the given spreadsheet and highlighting those records (on the map or as a recordset?

## Ideal Workflow

- Filter down to a set of records - option of removing select records
- Save the subset under a contract site list A
- Share that link with the hired contractor
- Contractor will be able to turn the layer on - list of sites would show up as a layer

## Goals
    - [ ] Should be able to filter records then Add/Remove Items from the List as list. Not locked to the filters. 
    - New tab? New Recordset Type?

## Initial Table Idea

Add two new tables, One for SiteLists, and a composite table for mapping Records to SiteLists. 
Image

## Additional Comments

- Currently we can Cache from a list of IAPP OR Activity Records using the Ids. This can be leveraged for the caching step.

### LINKED ISSUE #4098: :zap: determine registered users with `account_status`
# Describe the task

Fidx Renders on frontend to use `activated` determining membership instead of roles. 

## Value Add
- Consistency between ends
- not adding new states

## Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

## Additional Context

-

LocalNewsTV:
# Overview

- Set `loggedInOrWorkingOffline` to be based on account_status of user
    - account status only set when users are approved, so fills the void from using roles 
    - does not clobber state with token refresh.
- Closes #4098


meghna0593 (approved):

plasticviking (approved):
looks reasonable

PR #4103, opened 2025-05-20
[INV-3649] Merge 'Copy' and 'Paste' into 'Duplicate' records

### LINKED ISSUE #3649: Replace Copy/Paste with `Duplicate Record`
# Describe the task

The Copy/Paste functionality isn't intuitive, persists a record in state, and only really works if you paste into a record of the same type, but this context isn't visible anywhere, so you can paste into the incorrect record type, causing half the values to be irrelevant and hidden. 

The goal of the functionality is to duplicate an existing record, so the workflow should be simplified to just making a new record.

## Value Add
- Reduced Cognitive load
- Clear call to action
- Less juggling state / function calls

## Acceptance Criteria

- [ ] Remove Copy / Paste functions
- [ ] Create button 'Duplicate this Record'
- [ ] When a user selects duplicate record, create a new copy of the record and go to that activity.

## Additional Context

-

LocalNewsTV:
# Overview

- Combines Copy and Paste into one button.
- Reworks `NewActivityDialog` Component to hold two states.
- Converts Dialog actions into action creators
- Fixes small bug for `isDisabled` useEffect where duplicating ones own record from cache keeps it disabled
- Closes #3649


meghna0593 (approved):
🚀

PR #4115, opened 2025-05-28
Refactor app startup and feature detection

plasticviking:
Primarily in support of Android, but should also assist with new feature development

Configuration that impacts app behaviour can come from 4 different sources, and all have been unified into a single interface. Configuration is loaded before the application is mounted, and will not change.

Configuration sources:

- build time config (ie target platform, mobile or web, debug build yes/no)
- run time config (environment-specific configs like the api service to use)
- determined at runtime based on device characteristics (chiefly memory for android, and eventually screen dimensions)
- overrides based on injected environment variables (for example, to develop a new feature like the componentized map).

`StartupCoordinator` is responsible for loading the configuration and features, detecting runtime capabilities, applying overrides, and starting the app.

Also sequentially runs startup tasks to ensure background services are up and running in a definite order to prevent startup race conditions. startup tasks may optionally publish data into a global react context.

adds `FeatureGated` element to easily restrict access to features.

Tests Android device memory on startup and selectively disables features.


app/src/UI/Predicates/FeatureGated.tsx
```diff
@@ -0,0 +1,19 @@
+import { PropsWithChildren } from 'react';
+import { useSelector } from '../../utils/use_selector';
+import { FeatureFlags } from '../../state/configuration/feature-flags';
```
LocalNewsTV:
```suggestion
import { useSelector } from 'utils/use_selector';
import { FeatureFlags } from 'state/configuration/feature-flags';
```

app/src/UI/StartupCoordinator/StartupCoordinator.tsx
LocalNewsTV, line 1: Leftover Logs in file

LocalNewsTV (commented):
- Pushed in commit that fixes the tests broken by new config setup (where it was being mocked)
- Getting `cache not available` at `SQLiteRecordCacheService.listRepositories()` in function `handle_USER_SETTINGS_GET_INITIAL_STATE_REQUEST. This occurs in both VM and physical device
    - Doesn't seem to be an issue in browser, where localForage is used. 
- Still not operational on physical 4gb tablet without crashing
- Inconsistency in accessing Mobile. Is it preferred that we use useSelectors and yields to access the static state of MOBILE, or should we be importing the buildTimeConfig?

app/src/state/configuration/feature-flags.ts
```diff
@@ -0,0 +1,106 @@
...
+      // disable intensive features on all but the most capable hardware
+      COMPUTED_FEATURES.MAP_DATABC_LAYERS.enabled = false;
+      COMPUTED_FEATURES.MAP_BAKED_RASTER_TILES.enabled = false;
+      COMPUTED_FEATURES.DEGRADED_EXPERIENCE_WARNING.enabled = true;
+    }
+    if (totalMemoryGB < 4 || lowVMMemory) {
+      console.warn(
+        'Device is too small to support the full experience, operating with some features disabled (>=4GB features off)'
+      );
+      // many of our users will be using a device of approximately this size
```
meghna0593, line 79: are we disabling caching on 4gb devices?

app/src/UI/Predicates/FeatureGated.tsx
plasticviking: good catch. my ide was misconfigured.

app/src/state/configuration/feature-flags.ts
plasticviking: Haven't figured out the right features yet, this was mainly there as a starting point. I think we'll probably end up disabling one or both of the caches on these small devices though.

PR #4133, opened 2025-06-03
[INV-3703] location search/plotting

### LINKED ISSUE #3703: Search location name/coordinates
A search bar on the map where you can type an address or place name or coordinates and the map will pan to that location and mark it on the map. There might be another ticket for this already, but I can't find it.

LocalNewsTV:
# Overview

## Frontend
- Create new Component to query address, or place marker at coordinate
    - Debounces to avoid spamming 
- Add Tests for Component 
- Add margin to `Draw mode` box to avoid crushing
- Refactor `DisplayComposite` to allow for horizontal scaling
    - also fixes boxes overflowing out 
- Add Action Creator to just place a marker at a location hooking into existing 'userHovered'

## API
- Create new endpoint to send addresses based on partial search strings `/address-search`
- Hooks into the [BC Geocoder API ](https://www2.gov.bc.ca/gov/content/data/geographic-data-services/location-services/geocoder)
- Add env `GEOCODER_API_BASE`. The Url is not a secret and is publicly available. use `https://geocoder.api.gov.bc.ca/addresses.json` 
- add env `GEOCODER_API_KEY` Presently works without it needed, could change. 

- Closes #3703
### Addresses
image
image

### Coordinates

image
image

> [!NOTE]
> Suggestions float above the below components, they don't push them down.

### The terrible scenario where you want everything open at once
image

### DIFF
--- a/api/src/paths/address-search.ts
+++ b/api/src/paths/address-search.ts
@@ -1,0 +1,134 @@
+import axios from 'axios';
+import { RequestHandler } from 'express';
+import { IParsedAddress } from 'sharedAPI/src/interfaces/IParsedAddress';
+import { Operation } from 'express-openapi';
+import { ALL_ROLES, SECURITY_ON } from 'constants/misc';
+import { getLogger } from 'utils/logger';
+import { InvasivesRequest } from 'utils/auth-utils';
+
+const NAMESPACE = 'address-search';
+
+const defaultLog = getLogger(NAMESPACE);
+const GET: Operation = [getHandler()];
+
+GET.apiDoc = {
+  description: 'Partial Address Lookup via BC Geocoder API',
+  tags: [NAMESPACE],
+  security: SECURITY_ON ? [{ Bearer: ALL_ROLES }] : [],
+  parameters: [
+    {
+      in: 'query',
+      name: 'addr',
+      required: true,
+      description: 'Partial address for lookup',
+      example: '123 Main Street'
+    }
+  ],
+  responses: {
+    200: {
+      description: 'Results of search',
+      content: {
+        'application/json': {
+          schema: {
+            type: 'object',
+            properties: {
+              request: {
+                type: 'string',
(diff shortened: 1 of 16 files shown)


api/src/paths/address-search.ts
```diff
@@ -0,0 +1,137 @@
...
+        minScore: '45',
+        exactSpelling: 'false',
+        autoComplete: 'true',
+        addressString: addrString,
+        matchPrecision: 'block,street,locality'
+      });
+
+      const BASE_URL = process.env.GEOCODER_API_BASE;
+
+      const { data } = await axios.get(BASE_URL + '?' + params, { headers: { 'x-api-key': GEOCODER_API_KEY } });
```
plasticviking: any chance we can prefer `fetch` over `axios`? we're not completely there yet, but I'm trying to deprecate that dependency

sharedAPI/src/openapi/api-doc/api-doc.ts
```diff
@@ -90,7 +90,7 @@ export const api_doc = {
   ],
   externalDocs: {
     description: 'Visit GitHub to find out more about this API',
-    url: 'https://github.com/bcgov/lucy-web.git'
+    url: 'https://github.com/bcgov/invasivesbc'
```
plasticviking, line 93: 🤣

plasticviking (commented):
looks good. can you share test geocoder keys oob?

api/src/paths/address-search.ts
LocalNewsTV: Refactored to use `node-fetch` until we update our images. Then the import can be deleted (along with the dependency from package.json) and it will continue to work

meghna0593:
Looks good! 
Typo in the title - Add env  ~~`GEOCODER_BASE_URL`~~ `GEOCODER_API_BASE`

meghna0593 (approved):

PR #4136, opened 2025-06-04
[INV-4017] Add User Guide

### LINKED ISSUE #4017: Create In App User Reference Manual
# Describe the task

- Create new Page in application as a quick user reference guide, so Users can get information about a given feature in the application.

## Value Add
- The User manual can be taken offline, so users in the field can access it with the latest information
- Updating entries is easy and consistent, and doesn't rely on creating new PDFs or manuals, uploading them and updating links
    - Version control on the manual 
- App information is kept in the same place as our code
- We can add Search/Filter functionality so users can find what they need and fast

## Acceptance Criteria

- [ ] New Page created `User Manual` [name pending]
- [ ] User Manual has filter functionality so users can find what they need
- [ ] User Manual can display Icons and Images making it easier to use and reference
- [ ] User Manual uses Objects to populate entries, so new pages can be updated easily and consistently without need for new components (Like how `Whats New` works)

LocalNewsTV:
# Overview

> [!NOTE]
> Added Flags to control rendering until @brennanwebster  adds contents. Making Ticket made to remove flags after.

## Changes
- Add Component + Subcomponent to have User guide information for troubleshooting tips and information when on Mobile


app/src/UI/Features/UserGuide/guideEntries.ts
```diff
@@ -0,0 +1,21 @@
+import UserGuideEntry from 'interfaces/UserGuideEntry';
+
+const guideEntries: Array<UserGuideEntry> = [
+  // {
+  //   title: '',
+  //   titleIcon: '',
+  //   content: [
```
plasticviking: Is there a compelling reason not to just use JSXElement or ReactNode for the type of `content`?

app/src/UI/Features/UserGuide/faqEntries.ts
```diff
@@ -0,0 +1,21 @@
+import UserGuideEntry from 'interfaces/UserGuideEntry';
+
+const faqEntries: Array<UserGuideEntry> = [
+  // {
+  //   title: '',
+  //   titleIcon: '',
+  //   content: [
```
plasticviking: Is there a compelling reason not to just use JSXElement or ReactNode for the type of content?

PR #4175, opened 2025-06-25
[INV-4149] Uploadable SiteLists

### LINKED ISSUE #4149: Being able to upload a Sitelist excel sheet in IBC (Phase. 1)
# Describe the task

Allow a user to upload a Site list to the IBC. 

## Value Add
Bridging the Excel Sheet and InvasivesBC while the rest is being created

## Acceptance Criteria

- [ ] As a User, I want to Upload a Site list to InvasivesBC, and have access to that information from the App
- [ ] As a User, when I upload a Site List, I want an IAPP/Activty Recordset to be created.

LocalNewsTV:
# Overview

> [!IMPORTANT]
> Requirements Changed Midway through development so some changes may seem out of place (action creators for map boundaries) but were still improvements

## Front

- Allow a user to upload a supplied site list, and generate an IAPP and Activity Recordset
    - Add new object key `ids_to_filter` to limit the cap of the recordset to only these records before applying filters
         - This key was originally used for `Whats Here` id slice functionality. so modications were made to extend its usability
    - Modify Caching and `query` functions to handle ids_to_filter keys when used.
    - Site List modal accepts and parses Excel documents, displaying specific errors if applicable (missing fields, empty cells)
        - Bad IDs are filtered out on upload.
- Change actions regarding Client Boundaries from Strings to actionCreators
- Create a Plug and play Modal with some predefined CSS (permitting use of three specific classes in the contents `.header`, `.content`, `.control`
    - This modal isn't everything for everybody, and will have its CSS adjusted approriately as it gets implemented
    - This was made because there's several modals in the app all styled nearly identically given a few pixels of indifference
    - Setup in a way where we can add custom colour palettes as needs change

## Back

- Refactor SQL queries to use SiteLists when provided (IAPP, IBC, Batch Requests)
 - Verify all Ids brought in this way with Regex
 - Modify to allow Short IDs in addition to Long Ids 

## Shared

- Create common Regex file

## Packages
- Add XLSX package for Excel parsing

- Closes #4149

### New button in Records

image

### Initial Modal

image

### No data
image

### Needed column missing
image

### Some fields empty
image

### Result
image


meghna0593 (changes requested):
Tested out the functionality. Overall works as expected. Switched ID to SiteID and uploaded the file -- validations work too. 

One thing to fix would be the Activity recordset for the SiteList. Caching the recordset and going offline doesn't work as expected. The rows show up for a fraction of a second and disappear. The pagination is intact, however the records itself disappear. The toggling of the layer works too. 

Screenshot 2025-06-26 at 11 33 09 AM

meghna0593 (approved):
Looks good!!

PR #4181, opened 2025-07-04
Enable Geometry Editing When Geo-Tracking Is Paused or Self-Intersection Occurs

### LINKED ISSUE #4164: Pause / Resume for Geo-tracking should work as expected
Changes include:

- [x] 1. Enable shape editing while Geo Tracking is paused
   - Allow users to modify the drawn shape when geo tracking is paused, without needing to end the session.
- [x] 2. Allow editing of self-intersecting shapes during Geo Tracking -- 
   - Currently, geo tracking stops and the shape is closed when a self-intersection is detected. This change will allow users to pause, edit the shape to resolve the intersection, and then continue geo tracking without interruption.

meghna0593:
closes #4164

### DIFF
--- a/app/src/state/sagas/activity.ts
+++ b/app/src/state/sagas/activity.ts
@@ -352,6 +352,7 @@ function* handle_MAP_TOGGLE_TRACK_ME_DRAW_GEO_RESUME() {
 }
 
 function* handle_MAP_TOGGLE_TRACK_ME_DRAW_GEO_PAUSE() {
+  yield put(Alerts.create(mappingAlertMessages.canEditInfo));
   yield put(Alerts.create(mappingAlertMessages.trackingPaused));
 }
 
--- a/app/src/state/sagas/activity.ts
+++ b/app/src/state/sagas/activity.ts
@@ -271,9 +271,26 @@ function* handle_MAP_TOGGLE_TRACK_ME_DRAW_GEO_STOP() {
     console.error(err);
   }
   if (geographyWillContainIntersections) {
-    yield put(GeoTracking.exit());
-    yield put({ type: ACTIVITY_UPDATE_GEO_REQUEST, payload: { geometry: [newGeo] } });
-    yield put(Alerts.create(mappingAlertMessages.trackMyPathStoppedEarly));
+    const callback = (userConfirmsExit: boolean) => {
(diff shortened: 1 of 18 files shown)


plasticviking:
looks good and tests ok to me

app/src/state/sagas/activity.ts
```diff
@@ -275,9 +271,26 @@ function* handle_MAP_TOGGLE_TRACK_ME_DRAW_GEO_STOP() {
...
+          GeoTracking.exit(),
+          { type: ACTIVITY_UPDATE_GEO_REQUEST, payload: { geometry: [] } },
+          Alerts.create(mappingAlertMessages.trackMyPathStoppedEarly)
+        ];
+      }
+    };
+    yield put(
+      Prompt.confirmation({
+        title: 'Errors in current geography',
+        prompt: `You've attempted to stop tracking, but self-intersection exists, do you want to abandon your progress?`,
```
LocalNewsTV:
```suggestion
        prompt: `You've attempted to stop tracking, but the shape intersects itself. Do you want to discard your progress?`,
```

The `but self-intersection exists` reads a bit awkwardly.

LocalNewsTV (commented):
Works as expected, no unanticipated errors or logs.
- One minor text edit request.

One improvement suggestion:
If a shape has an issue at time of pause (It's already red), can we default into editing mode? 
It's eliminates a step in the flow `Pause` -> `Edit` -> `Actual Editing` -> `Save`.
This would reduce cognitive load of the worker who is already out in the field with the (hopefully) sun on their backs

meghna0593:
> Works as expected, no unanticipated errors or logs.
> 
> * One minor text edit request.
> 
> One improvement suggestion: If a shape has an issue at time of pause (It's already red), can we default into editing mode? It's eliminates a step in the flow `Pause` -> `Edit` -> `Actual Editing` -> `Save`. This would reduce cognitive load of the worker who is already out in the field with the (hopefully) sun on their backs

The functionality was intended to pause when an intersection occured. A piece of logic that handled this was removed, and I overlooked the resulting behavior. It should now correctly pause when a self-intersection is detected. Thanks for pointing that out.

LocalNewsTV (approved):

PR #4206, opened 2025-07-16
Layer effect

plasticviking:
# Overview

- {List all the changes, if possible add the linked issue/ticket #}

## Type of change

- [ ] Version change

## How Has This Been Tested?

## Screenshots

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code


app/src/UI/Features/LegacyMap/Controls/PrimaryLayerSelect.tsx
```diff
@@ -1,21 +1,20 @@
...
-  const dispatch = useDispatch();
+type PrimaryLayerSelectProps = {
+  layers: InvasivesMapLayerDefinitionWithState[];
+  selectLayer: (layer: string) => void;
+};
 
-  const { baseMapLayer, availableBaseMapLayers } = useSelector((state) => state.Map);
-  const offlineDefinitions = useSelector((state) => state.TileCache?.mapSpecifications) ?? [];
+const PrimaryLayerSelect = ({ layers, selectLayer }: PrimaryLayerSelectProps) => {
+  // const offlineDefinitions = useSelector((state) => state.TileCache?.mapSpecifications) ?? [];
```
LocalNewsTV: crumbs

app/src/UI/Features/LegacyMap/LayerPicker/LpLayers/LpLayers.tsx
```diff
@@ -14,10 +16,16 @@ const EmptyCollection = ({ text }: EmptyListProps) => (
...
+
+type PropTypes = {
+  layers: InvasivesMapLayerDefinitionWithState[];
+  setOverlayState: (layer: string) => void;
+};
+
+const LpLayers = ({ layers, setOverlayState }: PropTypes) => {
+  // const handleWmsClick = (layer) => {
+  //   dispatch({ type: TOGGLE_WMS_LAYER, payload: { layer } });
+  // };
```
LocalNewsTV: crumbs

app/src/UI/Features/LegacyMap/LayerPicker/LayerPicker.tsx
```diff
@@ -88,20 +95,20 @@ export const LayerPicker = () => {
                     <li>
                       <hr />
                     </li>
-                    {MOBILE && (
+                    <FeatureGated requires={'CACHE_TILES'}>
```
LocalNewsTV, line 101: One was changed to a `CACHE_TILES` feature gate, the other is still `MOBILE`, so two different conditions depending on LP State

app/src/UI/Features/LegacyMap/helpers/functional/layers-hook.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (
+      availableLayerDefinitions.filter((l) => l.name === 'Public-Vector' && l.active == !loggedInOrWorkingOffline)
+        .length > 0
+    )
+      return;
+
+    const updatedLayerDefinitions: InvasivesMapLayerDefinitionWithState[] = [];
+
+    for (const l of availableLayerDefinitions) {
+      const updated = l;
```
LocalNewsTV:
```suggestion
      const updated = { ...l };
```
This seems to correct the issue of immutability. good 'ol memory references

app/src/UI/Features/LegacyMap/helpers/functional/layer-definitions/baked-vector-simplified.ts
LocalNewsTV, line 1: image
Stale bread, or part of one of the branches to be merged in?

LocalNewsTV (commented):
This immutability error occurs often on load. first load on Mobile it happened 100% of the time. Could circumvent it by loading web mode, loading page, then changing to mobile mode. Performing a hard refresh caused the error to happen again.
image

Error results in a blue screen. Tend to notice it more in the web mobile version than the web version, though it happens in both.

**WMS Layers are not working when enabled. On Mobile/Web they result in a 400 error**
image

**On Android, they result in this error**
image

When comparing the request between Dev/Local I notice that the `bbox` key isn't populating the same (left: Local, right: Dev)
image

WMS layers no longer persist on a page refresh, or when going between online -> Offline. I expect the end users to find the second case especially annoying if they have a brief connection outtage. 

Draw tool dots don't show up on Map, these layer definitions are seemingly not-present.
Draw tools themselves work, the edit functionality works, but all visual indicators are gone. 

- `handle_USER_SETTINGS_GET_INITIAL_STATE_REQUEST` makes a `listRepositories` request that consistently crashes on my physical android device. During testing this only occurred on my physical android device (4GB) and didn't occur on a simulated 8GB, or Web

## TL;DR
- Unable to properly test WMS Layers due to bad requests.
- No visual indicators for draw tools
- Suggestion left that seems to fix immutability error causing blue screen crash.
- App feels more peformant from the state cleanup/flow-control
- Issue with Recordset DB being ready.

LocalNewsTV (changes requested):
> [!NOTE]
> Not everything in the Review might be completely related to the PR. But this is the first PR Where the physical Android devices have been able to run the application without crashing, and so other non-related issues might be mentioned.
> Some issues might be mentioned multiple times if they were consistent across devices.

# Code

Nothing stands out as egregious that I can pinpoint/target, but there are issues in the flow control.

# [Web Testing]

On first entry into the app (Cleared App Cache/Private Browser) The base map layer buttons and map attributions rapidly fire between two states until one is selected by the user.
- This only occurs on Web, does not occur on Mobile
- Once a field is selected, the issue doesn’t re-occur due to the persistence

Offline Map Tiles Layer Picker Option available in Web when shouldnt be

Base map option buttons don't appear when offline

image

A useEffect doesn’t  stop firing, resulting Ian a `maximum update depth exceeded` error.
- This only happens in Web, and not in Mobile

 image

This error gets thrown but both components continue to work as intended. You can get this by refreshing on the `/Records` page.
- This issue does not happen when running `Mobile`
- 
image

WMS Layers

- Persist on refresh and toggling Offline/Online.
- WMS Layers are working as expected
- When Offline, the Unavailable text no longer renders 

image

Base Maps

- BaseMaps are working as 
- BaseMaps choice persists between loads

Draw Tools
- Now behave and show as expected


# [iOS Testing]

WMS Layers
- WMS Layers not working, requests come back like this
image
- Notice this error a lot, not sure if its consequences  
image

Base Maps
-  Base Map Options don't appear when offline. 
    - Gray maps is the default, does not default into the baked tiles.

Offline Maps
- Can Download Offline Maps
- Can Enable Offline Maps from Layer Picker
    - Cannot enable offline maps from /OfflineMaps
  
# [Android Testing] - Physical Lenovo 4GB Ram Device

>[!IMPORTANT]
> Android didn’t crash a single time during testing :)
>App **did not** crash from going into offline mode and **not** having baked map tiles (Watching the Resource Monitor did not result in a climb of CPU). Previously this was a fast crash.

> Testing exclusively on the Normal layout, not that decoupled one for dev.

Base Maps
- Base Map options not available when offline
WMS Layers
	- WMS Layers not available as option

Overlay Behaviour
- Initially changing the Overlay positioning did not work. Touchees and Drags resulted in a downward drag of roughly 50px
    - The Behaviour feels like its keeping the overlay handler as the active target, so the next tap continues this issue, and the one following behaves as expected
- Oddly, after 15 or so minutes of testing/comparing results, the Overlay began to behave as expected
    - After this happened, the items in the Overlay no longer scrolled. Items visible can be interacted with as expected, the drag event is getting consumed.

Offline Maps
- Was able to download Offline Map tiles on device
- Layer Picker did not show Offline Map Tiles
- Attempting to enable Offline Map Tiles from /OfflineMaps resulted in nothing happening
    - Toggling Recordsets still worked, so this is not an `icon button` issue
- Inconsistency with the Feature Flags around this?

# [Android Testing] - 8GB Simulator

Same Overlay behaviour as before. Didn't get the Overlay to work though, but still not sure how I triggered it the first time.
- Sometimes Dragging the Map will also move the Overlay.

WMS Layers
- WMS Layers Available to select
- WMS Layers do not work when toggled on

BaseTile Options not available when Offline

Offline Maps
- Can Download Offline Maps
- Can Enable Offline Maps from Layer Picker
- Cannot Enable Offline Maps from /OfflineMaps page

## [Majors Summary]
- When Offline, BaseMap options are not available, and it defaults into gray maps. This makes the offline functionality unusable in current form
- Offline Map Tiles are being initialized in the /OfflineMaps section, so the Cached Map Tiles area of the Layer Picker Doesn't get populated until you've been on the /OfflineMaps page.


## [Concerns]
This branch is not in sync with dev currently. I am unable to test if the download continuity of Offline Maps is still in working order. 

## [Other Notes]

It seems the Layer Picker is firing a different action than the Offline Maps page, and is the likely culpriot
LP - `togglePreferredOverlayLayer`
Maps Page - `toggleOverlay`

image

LocalNewsTV (commented):
## iOS

- Baked tiles non-selectable when offline. This would make app undeployable.
- [Fixed] WMS Layers are now working
- [Fixed] Caching / Displaying working
-  Map tile continuity appears to work as expected

## Android - 4gb

- Still not crashing
- WMS doesn't display with a weird render of no-table, because it doesnt render at all
- Drag issue still present but not part of this branch
- Can download offline maps but not use them. Neglible since we don't have a full release
- Baked Maps don't work

## Android 8gb Sim

- [?] Unsure if WMS Layers work due to API changes and the simulator won't take my ENV changes to point locally
- Can Download/enable maps
- Also doesn't crash

## Web
- [Fixed] Tile Picker availability removed 
- [Fixed] useEffect chain no longer firing


## Bonus
- [Fixed] Toggling off Offline Map caches works now. This is an existing problem in dev

LocalNewsTV (dismissed):

LocalNewsTV (approved):

PR #4208, opened 2025-07-22
fix slow filter shape queries and make map layers visible

brennanwebster:
This fixes the slow queries that have been stacking up in prod when a user uses the whats here on a large area or uses a filter shape. Also makes the map layers visible when they are enabled.


LocalNewsTV (approved):
- Permission filters still work
- Permission rules still send correct
- Significant speed improvements
- Vector tile layers work as expected now

Issue #4229, opened 2025-08-05
Mobile: App loads slowly on initial login after creating 8+ custom recordsets


meghna0593:
When tested on an iPad, the app experiences significant slowdown during initial login if 8 or more custom recordsets were created in a previous session.

Steps to reproduce:
1. Create 8 or more custom recordsets.
2. Close the app.
3. Reopen the app and log in.

Expected Behavior: The app should load promptly regardless of the number of custom recordsets.

Actual Behavior: The app takes noticeably longer to load when initializing with 8+ custom recordsets.

PR #4230, opened 2025-08-05
Fix crash when adding more than 4 custom recordsets

### LINKED ISSUE #4219: Mobile: Adding 4+ custom recordsets causes crash
I've had varying results after testing this issue a few times. The first time the app crashed on me it was after adding my 4th custom recordset. The next few times it was between 6 and 8 added recordsets. It seems like adding the recordsets causes a memory usage spike and eventually crashes the app. I don't think the app needs to handle 8 recordsets necessarily but 4 or 5 is probably a reasonable amount.

meghna0593:
# Overview

- Fixes #4219 
- Tested on the browser, iPad and iPad simulator
- Noticed that the app is slow on initial load after creating 8+ recordsets.  [#4229](https://github.com/bcgov/invasivesbc/issues/4229)

## Type of change

- [ ] Version change

### DIFF
--- a/app/src/state/sagas/map.ts
+++ b/app/src/state/sagas/map.ts
@@ -547,12 +547,12 @@ function* handle_MAP_INIT_FOR_RECORDSETS() {
   const recordSets = Object.keys(userSettingsState.recordSets);
 
   // current layers
-  const layers = yield select((state) => state.Map.layers);
+  const mapState = yield select(selectMap);
+  const layers = mapState.layers;
   const layerIDs = layers.map((layer) => layer.recordSetID);
 
-  // current but unintialized:
   const currentUninitializedLayers = layers
-    .filter((layer) => !layer.loading)
+    .filter((layer) => !Object.prototype.hasOwnProperty.call(layer, 'loading'))
     .map((layer) => {
       return { recordSetID: layer.recordSetID, recordSetType: layer.type };
     });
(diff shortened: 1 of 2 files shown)


app/src/state/sagas/map.ts
```diff
@@ -547,12 +547,12 @@ function* handle_MAP_INIT_FOR_RECORDSETS() {
...
   // current layers
-  const layers = yield select((state) => state.Map.layers);
+  const mapState = yield select(selectMap);
+  const layers = mapState.layers;
   const layerIDs = layers.map((layer) => layer.recordSetID);
 
-  // current but unintialized:
   const currentUninitializedLayers = layers
-    .filter((layer) => !layer.loading)
+    .filter((layer) => !Object.prototype.hasOwnProperty.call(layer, 'loading'))
```
LocalNewsTV, line 555: is there a reason to use this prototype instead of ES2022 supported Object.hasOwn?

LocalNewsTV (dismissed):
Solution corrects the regression previously introduced.

app/src/state/sagas/map.ts
meghna0593: No specific reason — either approach works. The one I used is the older, widely supported format (Object.prototype.hasOwnProperty.call). The method you suggested (Object.hasOwn) is recommended for modern environments.

plasticviking: +1 for this. Also, I think this line is probably the only important one?

meghna0593: Yes, looks like this is the only change needed to fix the crashes (lol). 

However, there's a noticeable slowdown on initial login when there are more than 8 custom recordsets. Layer toggling behaves inconsistently on page refresh — sometimes it works, sometimes it doesn't.

PR #4243, opened 2025-08-13
Add Platform Tag for New Records

### LINKED ISSUE #4216: Mobile: Add tag to differentiate web vs mobile created records
Maybe using the sync_status column in the main table, I don't think its used for anything although it might be good to check.

meghna0593:
# Overview

- This PR adds a new column `platform_src` to the activity table to track the platform used when creating an activity.
- Verified data insertion from both Web and iPad platforms.
- Closes #4216 

## Type of change

- [ ] Version change

To test:
- Run database migrations.
- Create and submit a new activity.
- Verify that the platform_src column is correctly populated in the activity table.

### DIFF
--- a/api/src/models/activity.ts
+++ b/api/src/models/activity.ts
@@ -44,6 +44,7 @@ export class ActivityPostRequestBody {
   species_treated: string[];
 
   jurisdiction: string[];
+  platform_src: string;
 
   /**
    * Creates an instance of ActivityPostRequestBody.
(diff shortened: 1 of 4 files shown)


LocalNewsTV (approved):
Creating new records works great, both iOS and Web platforms delivered the expected value when creating a new record, and the migration ran up and down with no issues. 

I did not test on Android since I expect no difference given its all tied to platform variable.

One consistency issue noted is that legacy records will appear as null, but updating them will appear as an empty string based on `activity.ts:103`, for consistency maybe we want null also.


image

*[empty string vs null, invasive plant column to show the row exists]*

If the record was already defined (created post-update), this doesn't occur (as expected)

image

*[iOS -> iOS]*

PR #4259, opened 2025-08-26
[INV-4256] Display Features on map click

### LINKED ISSUE #4256: Allow users to click on polygons on the map and see basic info
Currently users use the Whats Here button for this, but it would be simpler if they could just click directly on a polygon and see some info. Bonus points for clicking on an area of stacked records and seeing the info for all of them in a popup, and then being able to click the popup info to be taken to the record. Sort of a Whats Here lite, can discuss more if needed.

LocalNewsTV:
# Overview

- Create two new Components
    - *LayerDataMarker.tsx* `Controller` for the Popover. holds event listeners
    - *LayerDataMarkerContent.tsx* `View` for Popover.
- Modify Offline Activities layers to imitate tiles in vector layer. 
- Move some CSS colour values into app.css, refactor accordingly.
- Emit custom event from Edit Button

Add new feature that shows metadata for Recordset vectors with a single click. Users will be able to see the type of record, ID, and Map Symbols (pulled from current vector data). Users can also open a record quickly this way.

The Popover will not interrupt active map work, such as using `Whats Here` or drawing any shapes with the draw tools.
The Popover appears when at zoom level 12 or greater

> [!IMPORTANT]
> Currently this feature only works while online. Since the geojson work is being replaced very soon with PMTiles, it will be revisited for feasability after that is implemented 

Since maplibre doesn't have a native 'touch' event, and instead has start/end ones. the events mock a touch by using a `time between` on the two events. ensuring regular device usage doesn't get interrupted by popups.

- Closes #4256

## Screenshots

image

[Feature highlighting ]

image

[Many features in result]

### DIFF
--- a/app/src/UI/App.css
+++ b/app/src/UI/App.css
@@ -96,8 +96,14 @@ body {
   --success-green: #357a38;
   --warning-orange: #ffc107;
   --bc-yellow: #fcba19;
-  --blue-primary: #2196f3;
+  --blue-primary: #1976d2;
   --eigengrau: #16161d;
+  --popover-list-bg-color: #1e5189;
+
+  /* List Item Colors */
+  --even-list-item-bg-color: #eceae8;
+  --odd-list-item-bg-color: #fff;
+  --hovered-list-item-bg-color: #d1cfcd;
 }
 
 /* Color Classes */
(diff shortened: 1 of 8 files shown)


meghna0593:
Works well for the most part! 

A behaviour that could throw the user off -- when opening an already submitted activity and clicking on the polygon to edit its shape, the popup keeps appearing repeatedly. This interrupts the editing flow and could be frustrating to the user. 

It might be worth suppressing the popup in this context.

Screenshot 2025-08-26 at 4 05 19 PM
Screenshot 2025-08-26 at 4 03 17 PM

LocalNewsTV:
@meghna0593 
Good catch. Looks like the `edit` button isn't emiting the `draw.modechange` event. The WhatsHere button was the same, I'll look look for a fix in the morning.

meghna0593 (approved):
LGTM!

PR #4264, opened 2025-08-28
IOS Plugin for Offline PMTiles Support

plasticviking:
# Overview

This is one of several related pieces of functionality aimed at bringing better support for offline maps into the app. This provides a native IOS plugin to allow downloading of `pmtiles` maps, and a matching PMTiles-compatible `Source` implementation to load specific byte ranges from the native code into the pmtiles layer.

To test this code: Open the `Offline Proto Maps` debug pane from the zap menu (enable VITE_DEBUG on `build:ios` first by editing package.json), click the `Request Download` button, which will queue a native-code download of the public pmtiles file (as an example, for now). Progress will be reported in the dialogue. To see this on the map and see the driver working, edit `demo-offline-vector.ts` to reference the correct filename and rebuild the app. You should see a map equivalent to the public map, but coloured red to differentiate it.

Dynamically adding layers is not yet supported. Localforage code is stub-only for now and may be removed if native metadata storage is sufficient.

IOS will continue to download files in the background, even if the screen is off. The app does not need to be in the foreground to complete downloads, and download completion is unrelated to user authentication state.

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-08-28 at 18 13 28
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-08-28 at 18 25 27


app/src/UI/StartupCoordinator/StartupCoordinator.tsx
```diff
@@ -48,6 +51,8 @@ async function StartupCoordinator() {
     }
   }
 
+  console.dir(providedContext);
```
LocalNewsTV: May wish to gate this with a `debug` flag if persisting for a while between branches.

LocalNewsTV (commented):
>[!TIP]
> Running Debug on a physical iPad doesn't work. The app just fails once logged in, so testing on Simulator is advised [This is an old issue].

- [Physical] Tested Downloading PMTiles and immediately closing application and shutting off screen. Download succeeded.
    - Tiles render on screen demonstrated in screenshot.
    - Due to being unable to login because debug woes, the public map layer appears over the PMTile layer. Can confirm the PMTile is there by zooming in since they briefly won't overlap during zoom
- Observed Issue in CLI build where build fails due to Object version being `70` and not matching anything. Referenced Online comments with similar issues. GH Issue linked in the CLI Fail logs. 

Looking forward to the other stages of this work.

PR #4270, opened 2025-09-02
[INV-4229][INV-4231][INV-4272] Simplify the getIds/vector workflow

### LINKED ISSUE #4229: Mobile: App loads slowly on initial login after creating 8+ custom recordsets

### LINKED ISSUE #4231: Web: Custom recordset layers don't appear on toggle after page reload
Seen this happen intermittently

### LINKED ISSUE #4272: Whats Here button not returning results when it should
This has been an ongoing issue with the Whats Here button being fairly inconsistent. In my experience refreshing the page will often make it start working again, but users have had issues with it since it was first created. This screenshot is from mobile but the same issue happens on web also. Basically you see records on the map, click the Whats Here button, draw the box over the records you see, and the Whats Here table will show you no results.

Image

LocalNewsTV:
# Overview

## Added
- Setup actionChannel for getting IDs to avoid API Spam. Reduces memory consumption at App load.
- Set height and width of body to 100% of view, and overflow hidden to remove the jarring effect over scrollbars appearing randomly with tooltips. (Especially noticeable when hovering in recordset)
- `getIdsForRecordset` function to simplify the getIdsFor... workflow. Only requires a recordset, figured out the typing and endpoints on its own.

## Refactored
- Consolidate Activity/IAPP getIdsForRecordset(Success) actions into one WhatsHere flow.
    - WhatsHere chosen as Wrapper, because the IDLists are only used in WhatsHere functionality.
    - Delete references to individual Items
- Refactor mapLayers to only have one path to creation. Removes many race conditions and other issues that led to map inconsistency (toggles being out of sync, colours not being correct, etc)
    - Single path sets colour, layer state, toggle state, keeping everything in sync at app load.
    - there was several paths to create this, all with different interpretations of layerState values. 
- Refactor OfflineActivities to use the regular Maplayer actions. Delete special state keys they had assigned
- Reorder components in `StartupCoordinator` to listen for hydration action
    - setup callback in persistor to fire auth actions only **after** rehydrating. Ensures Recordsets are correct.
- Minor type enhancement with Enums
- Remove bandaid in `RecordSetControl.tsx` that fired getID actions to counter the app inconsistency in recordset.
- Remove optional key in colour. Set default value to  defaultRecordsets.

## Closes

- Closes #4229 
    - ActionChannel setup to drop memory consumption, run recordsets synchronously, and setup maps separately. 
- Closes #4231
    - Actions fire after hydration, resulting in `all` recordsets getting their calls. Not just default + cached
- Closes #4272
    - IDLists should now be populated, which WhatsHere relies on in its currentState


app/src/state/actions/whatsHere/WhatsHere.ts
```diff
@@ -2,12 +2,25 @@ import { createAction } from '@reduxjs/toolkit';
 import { SortFilter } from 'interfaces/filterParams';
 import { RecordSetType } from 'interfaces/UserRecordSet';
 
+interface IGetIdsForRecordset {
```
plasticviking, line 5: stylistically, I prefer not prefix interfaces. There isn't really any advantage to doing so in Typescript

app/src/state/sagas/map.ts
```diff
@@ -648,6 +621,46 @@ function* handle_ACTIVITIES_TABLE_GET_ROWS_REQUEST(action) {
...
+    if (!userIsOffline) {
+      const ids = yield getIdsForRecordset({
+        recordSetType: currentState.recordSets[action.payload.recordSetID].recordSetType,
+        tableFilters: filterObject.tableFilters
+      });
+      yield put(WhatsHere.getIdsForRecordsetSuccess({ idList: ids, ...action.payload }));
+      return;
+    }
+  } catch (e) {
+    console.error('[handle_GET_RECORDSET_IDS]:', e);
```
plasticviking, line 643: Is there condition handled in the reducer?

app/src/UI/StartupCoordinator/StartupCoordinator.tsx
LocalNewsTV, line 1: Re-ordering to listen to rehydration action more appropriately.
Ordering taken from [redux-persist docs](https://www.npmjs.com/package/redux-persist)

app/src/state/actions/whatsHere/WhatsHere.ts
LocalNewsTV: @plasticviking 
Should we maintain a consistent suffix or prefix on all interfaces created for facilitating createAction\<interface\>?

Now that the Redux actions are largely completed I could write a ticket to move all the interfaces to the interfaces folder if preferred. There I can strip the leading `i`'s, and the import path will add better clarity to their origin.

We can probably keep the filename structure matching to the actions themselves. `UserSettings`, `MapActions`, `WhatsHere`, etc. Otherwise we'll be introducing a whole lot of monofiles, and there's more than a few extending into others.

app/src/state/sagas/map.ts
LocalNewsTV: I don't quite follow what you're saying in this comment.

But I reorganized this function and commented it to better illustrate its intents.

app/src/state/store.ts
LocalNewsTV, line 1: Loose actions in this file were firing before re hydration occurred, resulting in race conditions.

PR #4285, opened 2025-09-11
[INV-4286] Allow extending Redirect_URI path 

### LINKED ISSUE #4286: Redirect user to AccessRequest after logging in
# Describe the task

## Value Add

When user logs in via requesting access they'll be redirected to the page instead of back home

## Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

## Additional Context

-

LocalNewsTV:
# Overview

- Allow REDIRECT_URI Path to be extended from the base for sign-in requests.
   - Implementation is scalable if future examples needed to be implemented down the line.
- Does not affect Logging in on through mobile application
- Closes #4286
 
## Use Case

1. User enters the application for the first time
2. User selects 'Request Access'
3. User is prompted to Login
4. User Successfully Logs in and is redirected to the `Access Request` page

In the current version, the user would be redirected back to the home page, as if the click onto `Access Request` never happened.

If a user logs in normally, they'll be redirected to the base url as normal.

### DIFF
--- a/app/src/state/sagas/auth/keycloak.ts
+++ b/app/src/state/sagas/auth/keycloak.ts
@@ -1,6 +1,5 @@
 import { call, cancelled, delay, fork, put, select, takeLatest } from 'redux-saga/effects';
 import Keycloak, { KeycloakLogoutOptions } from 'keycloak-js';
-import { useNavigate } from 'react-router';
 import { AppConfig } from 'state/configuration/runtime-config';
 import { selectConfiguration } from 'state/reducers/configuration';
 import { selectAuth } from 'state/reducers/auth';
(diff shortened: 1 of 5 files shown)


app/src/state/sagas/auth/keycloak.ts
```diff
@@ -23,7 +22,7 @@ function* handleSigninRequest(action) {
 
   try {
     yield call(keycloakInstance.login, {
-      redirectUri: config.REDIRECT_URI,
+      redirectUri: config.REDIRECT_URI + (action.payload?.extendedRedirectUri || ''),
```
plasticviking, line 25: payload cannot be undfined (? is unneeded)

PR #4301, opened 2025-09-17
[INV-4296] Filter records by 'Most Recent Observation'

### LINKED ISSUE #4296: Enable 'Show only Most recent Observation' Checkbox
# Describe the task

add checkbox filter to recordsets to enable showing only the most recent observations

## Value Add
Users can find what recordsets are the latest for their plant type

## Acceptance Criteria

- [ ] Checkbox added
- [ ] Query modified to allow this join check
- [ ] Vectors and bounding boxes still work as expected

## Additional Context

-

LocalNewsTV:
# Overview

- Add new Checkbox Filter `Most Recent Observation`
    - Implement filter in API query 
- consolidate `tableFiltersHash` assignments to one reducer. 
- Decouple filters from Recordset component
- Add table filter property 'hidden' to not display filters in the list
    - Current use cases include Draft Recordsets having the draft filter, and filtering by Current observation
- Modify caching to set `ids_to_filter` when using the `most recent observations` filter
    - Helps with the constraints also presented  by server side shape filters
- remove `viewFilters` redux state/action.
- Closes #4296

## Screenshot

image

## Testing

Checked for the following Regressions

**For the new Filter:** 
Compared the actual results vs the expected results with @brennanwebster to ensure working properly

**For Regressions:**
- Checked Vectors Filtered/Updated when
    - Adding a recordset (with a value (most recent Observation is the only case of this))
    - Modifying a filter
    - Deleting a Filter

**For Mobile**
Created and cached two recordsets in localForage where the filters both had the same created_by filter
One has the `Most Recent Observations` filter. 

Comparing the results of these while offline to ensure they are what was expected.

Did not test this in SQLite, as the logic nearly identical


meghna0593 (approved):
Noticed a regression when toggling the layers on/off for the cached recordset — they no longer appear. I tested this on the dev environment and observed the same behaviour, so it is unrelated to this ticket.

LTGM 🚀

PR #4317, opened 2025-11-04
API refactor continued

LocalNewsTV:
# Overview

- Refactor auth-utils to use single transaction. Immensely reduce complexity
- modify package.json to include `--inspect` flag. Allowing for simpler debugger connections 
- Reduce Logger from printing all context to a reduced subset
- Logger throws Errors more neatly
- Remove several try/catches to allow falling into the globalErrorHandler, preventing errors from being squashes and losing meaning
- Modify globalErrorHandler to handle Objects based in from express-openapi (passes in apiSpec format, not as an error)
- incorporate new tools into several endpoint as Proof of concept, ranging from multi-request to single request

image

```
[2025-11-03 16:17:06][error][roles]: Error: Release called on client which has already been released to the pool.
    at throwOnDoubleRelease (/Users//api/node_modules/pg-pool/index.js:27:9)
    at Client.release (/api/node_modules/pg-pool/index.js:338:9)
    at QueryHandler.close (/api/src/utils/endpoints/QueryHandler.ts:39:22)
    at QueryHandler.query (/src/utils/endpoints/QueryHandler.ts:55:32)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async <anonymous> (src/paths/roles.ts:45:7)
    [getRoles]
```


api/src/paths/activity/{activityId}.ts
```diff
@@ -1,144 +1,103 @@
...
-  return async (req: InvasivesRequest, res, next) => {
-    if (req.authContext.roles.length === 0) {
-      res.status(401).json({ message: 'No Role for user' });
-    }
-    defaultLog.debug({ label: '{activityId}', message: 'getActivity', body: req.params });
-
-    const activityId = req.params.activityId;
-    let connection: PoolClient | undefined;
+  return async (req: InvasivesRequest, res: Response, next) => {
+    if (!verifyUserRole(GET.apiDoc, req)) return res.sendStatus(401);
```
plasticviking: This seems like it could be done in middleware?

api/src/paths/activity/{activityId}.ts
```diff
@@ -1,144 +1,103 @@
...
-      return res.status(500).json({
-        message: 'Unable to fetch activity.',
-        request: req.body,
-        error: error,
-        namespace: 'activity/{activityId}',
-        code: 500
-      });
+    } catch (e) {
+      logger.error(e);
+      return res.status(500).send('Unable to fetch activity.');
```
plasticviking: What happens if we omit this and let the catch-all handler take it? Is there any real advantage to this catch block?

api/src/utils/auth-utils.ts
```diff
@@ -49,162 +93,78 @@
...
-        user: null,
-        roles: [],
-        filterForSelectable: filterForSelectable
-      };
-
-      resolve();
-    });
-  }
 
+  const token = req?.header('Authorization')?.replace(/Bearer /, ''); // e.g. "Bearer abc123efg456"
```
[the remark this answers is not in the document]
LocalNewsTV: False Positive.

api/src/paths/activity/{activityId}.ts
LocalNewsTV: verifyUserRole has been implemented into the bearer security handler

api/src/paths/activity/{activityId}.ts
LocalNewsTV: Removing this and throwing an arbitrary error as a test successfully catches it in the global handler.

PR #4397, opened 2026-01-26
normalized database - openshift dev candidate 0

plasticviking:
# Overview

- Add Terrestrial Plant Observation subtype import
- Add flag to import command to restrict by subtype
- Squash the migrations again because of another schema-related issue
- Add db router to fix the schema issue
- Add rudimentary JWT-based authentication
- Support bypassing this authentication in tests with runtime settings override
- Adjust some models and their dependent tests and serializers
- react debug client: support authentication

## Known issues

- SpecificUseCode should go back to being a OneToMany, but should target a specific instance of Terrestrial/Aquatic Observation rather than the activity itself.


api-rework/client/src/activities/detail.tsx
```diff
@@ -31,19 +32,38 @@ const ActivitiesDetail: React.FC = () => {
 
   const [tab, setTab] = useState('django');
 
+  const { state: auth } = useContext(AuthContext);
+
   useEffect(() => {
     setLoading(true);
+    console.dir(auth);
```
LocalNewsTV, line 39: Leftover log

api-rework/client/src/client.tsx
```diff
@@ -1,20 +1,181 @@
...
+  }, []);
+
+  useEffect(() => {
+    /* check every 5 seconds to see if we are going to expire */
+    const id = setInterval(() => {
+      console.log('freshness check');
+      dispatch({ type: 'keep_token_fresh' });
+    }, 5000);
+    return () => clearInterval(id);
+  }, []);
```
LocalNewsTV, line 145: This is the source of the 5 second timeout fail.

> Uncaught ReferenceError: can't access lexical declaration 'dispatch' before initialization

api-rework/invasives/api/legacy_db/mappings/wells.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def add_well_information(new: Activity, old: LegacyActivity):
+    if old.activity_payload.form_data.activity_subtype_data.Well_Information:
+        for (
+            well
+        ) in old.activity_payload.form_data.activity_subtype_data.Well_Information:
+            if well.well_id == "No wells found":
+                logging.info(
+                    "Omitting empty well spec because of magic string well id = 'No wells found'"
```
LocalNewsTV, line 14: This line was firing near constantly during the ETL process. Just a comment, not sure if you have any prefilters leading to this point or just let it run after each iteration

LocalNewsTV (approved):
# General comments

- restrict_to_subtype appears to be working as intended. the ETL is successful on importing the observation records. Still awaiting geometry to be added before we can see its a complete Observation detail.  
- Our ETL should probably include the RISO/FLRNO Columns. Currently it doesn't and the details were added to the entry part of the form. Doing so would save us from making a *lot* of subsequent API calls for the ETL.
- Auth appears to be working on the happy path

# Issues
- The Login page for the client will disappear into an error after ~5 seconds of not touching the authenticate button
- Migration log seems to just fire for fun (picture enclosed) Will fire during tests, migrating on a fresh DB, etc

image

[*Fresh DB Migration*]

image

[*Running Test Suite*]

PR #4439, opened 2026-03-17
Suggested Treatment IDs

LocalNewsTV:
# Overview

- Add Online/Offline Workloads to fetching Suggested IDs
- Implement Serializer to display info
- Implement endpoint in Django.
    - Filter Searches based on bucket as we are no longer restricting linking to Treatments
        - Monitoring -> Chemical -> Observation -//-> Observations don't request
- Adjust linked_activities model (short_id -> label (we shouldnt save this, but the user needs it for display))


api-rework/invasives/api/viewsets/ids_within_bounds.py
```diff
@@ -0,0 +1,51 @@
+from django.contrib.gis.geos import GEOSGeometry
```
plasticviking, line 1: I'd much rather see this as a part of a larger, integrated search mechanism. We have this proliferation of endpoints on the node API and I'd rather just have `/activity/search` work properly.

PR #4467, opened 2026-04-01
Mock form submission, styling, error handling

LocalNewsTV:
# Overview

- Add mock submission url/viewset to facilitate development
    - Modify Submit action to target API (no offline support yet) 
- Re-style Record Metadata
- Update Employer/Agency filters to allow `Not Required` by default
- Add created_by as optional param.
- Update Direct Application calculation to Multiple Spray (same results, BW tested)
- Refactor Create Record Dialog to use RHF components
- Add try-catch to Recordset row requests to avoid crash if API fails.


plasticviking (approved):
lgtm

PR #4516, opened 2026-04-29
Patch Recordset Caching

### LINKED ISSUE #4514: User can Cache Records
# Describe the task

Refactor the Caching Mechanisms to that a user can Cache records from the new dataset

## Value Add
Existing functionality ported over to new dataset

## Acceptance Criteria

- [ ] User can save a recordset
- [ ] User can delete a recordset
- [ ] User can load a record from a recordset
- [ ] Everything works offline

## Additional Context

-

LocalNewsTV:
# Overview

- Add `appearance: none` to form inputs to prevent iOS styling. 
- Add some basic colors to `App.css`
- init constants file, add regex for UUID and shortID
- add Extended Row Serializer for caching purposes
- Modify Recordset-rows to return array of ID's, (old API was `{ activity_id: str }`)
- set selectColumns param to optional as its implicit to `all`
- modify getIdsByRecordset to handle both new/old payloads (because IAPP is legacy)
- Adjust Recordset Caching to
    - Target Django API
    - Expect Smaller payloads
    - consume already formatted payloads. 
- Closes #4514


api-rework/invasives/api/viewsets/recordset_rows.py
```diff
@@ -1,11 +1,26 @@
...
-from api.serializers.activity_recordset_row import ActivityRecordsetRowSerializer
+from api.serializers.activity_recordset_row import (
+    ActivityRecordsetRowSerializer,
+    CachedActivityRecordsetRowSerializer,
+)
 from api.models.activity import Activity
+from api.constants import uuid_regex, short_id_regex
+from invasivesbc.settings import LEGACY_DB_CONNECTION_STRING
+
+logging.basicConfig(level=logging.DEBUG)
```
plasticviking: should be set in settings module (or just on command line)

plasticviking:
Other than the loglevel clobber, it looks good to me

api-rework/invasives/api/viewsets/recordset_rows.py
LocalNewsTV: Thanks for the note. I was using the admin command implementation as my example. 
I've updated it to include an `invasives` logger, added an ENV for better control, and a formatter. 

Example output from a simple raised Exception. 
```
[ERROR][recordset_rows][2026-05-04 14:57:10,644]: Error while handling 'spatialFilterDrawn'
Traceback (most recent call last):
  File "/Users/.../recordset_rows.py", line 154, in _build_single_filter_q
    raise Exception("It was inevitable")
Exception: It was inevitable
```

PR #4540, opened 2026-05-14
[INV-4523] Implement Autofilled Activity values

### LINKED ISSUE #4523: Implement Computed Values for Incoming Records
# Describe the task

Implement functionality so the computed values are filled when new records are posted

## Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

## Additional Context

-

LocalNewsTV:
# Overview

- Activity Creation implements computed values with eventual consistency via `celery` 
- Update `RecordsetRowSerializer`with changes necessary to properly display the now-populated tables.
- Update activity create logic to prevent recursion
- Closes #4523

### DIFF
--- a/api-rework/invasives/api/serializers/activity_recordset_row.py
+++ b/api-rework/invasives/api/serializers/activity_recordset_row.py
@@ -2,7 +2,7 @@ import json
 from rest_framework import serializers
 from django.contrib.gis.db.models.functions import Centroid, AsGeoJSON
 from api.models.activity.activity import Activity
-from api.models.activity import ActivitySubtypes
+from api.models.activity import ActivitySubtypes, RisoArea, ProjectCode, FundingAgency
 from api.serializers.activity import ActivitySerializer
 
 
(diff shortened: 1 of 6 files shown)


api-rework/invasives/api/serializers/activity_recordset_row.py
```diff
@@ -236,6 +234,12 @@ def get_species_biocontrol_full(self, obj):
 
         return self.build_response_value(sorted(set(agents)))
 
+    def get_regional_invasive_species_organization_areas(self, obj):
+        risos = []
+        for record in obj.activitydatarecord_set.all():
```
plasticviking: is there some reason not to just return an array?

api-rework/invasives/api/tasks/generate_computed_activity_fields.py
```diff
@@ -0,0 +1,183 @@
...
+    if elevation is None:
+        return
+    a.computed_elevation_m = elevation
+
+    a.save(update_fields=["computed_elevation_m"])
+
+
+@shared_task(bind=True, max_retries=3, default_retry_delay=30)
+def generate_computed_activity_fields(self, record_id):
+    try:
```
plasticviking, line 151: since you call a.save() in the `fetch_`... handlers, it's probably best to wrap this whole function in a tx.begin()...commit() block to avoid the possibility of partial updates (django uses autocommit by default)

api-rework/invasives/api/serializers/activity_recordset_row.py
LocalNewsTV: For the frontend these only get used in the recordsets for display/filter purposes, so it moves compute to our resources making it friendlier for an end user (if even marginally)
But the bigger boon is when caching records, the device that is busy caching also does not need to start formatting this. 

Previously the frontend would use `getUnnestedFieldsForActivity` to parse every incoming record to format it nicely, that is now deprecated

<sub><i>It's also a marginally smaller payload, but that wasn't a consideration</i></sub>

api-rework/invasives/api/tasks/generate_computed_activity_fields.py
LocalNewsTV: I've reformatted the functionality to get the needed data beforehand in order to not have third party APIs become the bottleneck.

The creation/update logic has been moved into a transaction block as per your suggestion.

PR #4590, opened 2026-06-24
Add WKT Geometry tool

plasticviking:
# Overview

- Add WKT Batch Geometry tool
- Clean up some races/dependency issues in the main map

Not yet tested on IOS (applicable only to main map changes, since batch is not availble on IOS)

### DIFF
--- a/app/src/UI/Features/Batch/BatchGeometryTool.css
+++ b/app/src/UI/Features/Batch/BatchGeometryTool.css
@@ -27,11 +27,11 @@ div.geometry-tool {
     grid-area: help;
 
     p,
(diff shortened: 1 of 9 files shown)


plasticviking:
I have since tested out the map changes on IOS and it looks good to me.

LocalNewsTV (commented):
I opened a PR against this one for minor styling discrepancies, but some improvements that might be added: 
- Adding Layers for the draw tools so make linestrings and points more clearly visible
- Adding layers for the draw tools to see the vertexes

It's just matching the existing functionality we have in-app already.

Issue #4530, opened 2026-05-07
Batch enhancement: allow point uploads with an area between 1-10


brennanwebster:
Hopefully this is doable within the current templates, instead of having to create another set of point templates. Would need an extra column for Area that is used only for points. Area must be between 1-10 square meters. Probably should turn the point into a little octagon with the given area to be consistent with other things we've been planning.

PR #4591, opened 2026-06-24
Add Point Area to batch shared columns

plasticviking:
# Overview

- #4530

### DIFF
--- a/api/src/utils/batch/definitions.ts
+++ b/api/src/utils/batch/definitions.ts
@@ -24,6 +24,7 @@ export class TemplateColumn {
   helpText: string | null;
   required = false;
   overwritesPrevious = false;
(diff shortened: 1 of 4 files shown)


LocalNewsTV (approved):

api/src/utils/batch/definitions.ts
```diff
@@ -24,6 +24,7 @@ export class TemplateColumn {
   helpText: string | null;
   required = false;
   overwritesPrevious = false;
+  intentionallyUnmapped: boolean = false; // suppress object mapper warnings for columns with no mapping (eg Point Area)
```
LocalNewsTV, line 27: Records using old templates that do not have the `Point Area` column get denied as they are missing the column. Is this meant to work around that, or or is this for something else? :eyes:

PR #4683, opened 2026-09-09
Reduce Calls in computed fields celery task.

LocalNewsTV:
# Overview

- Refactor BCGW Calls into one request.
    - Cuts 4 calls into 1 
- Refactor `fetch_computed_invasive_plant_management_areas` and `fetch_computed_regional_districts` into `query_singleton_spatial_tables`
    - Cuts down one database connection per rotation


api-rework/invasives/api/tasks/generate_computed_activity_fields.py
```diff
@@ -125,49 +174,25 @@ def fetch_computed_regional_districts(a: Activity):
...
-        a.computed_elevation_m = fetch_computed_elevation_m(a)
-        a.computed_biogeoclimatic_zone = fetch_computed_biogeoclimatic_zones(a)
-        a.computed_flrno_districts = fetch_computed_flrno_districts(a)
-        a.computed_invasive_plant_management_areas = (
-            fetch_computed_invasive_plant_management_areas(a)
-        )
-        a.computed_moti_districts = fetch_computed_moti_districts(a)
-        a.computed_ownership = fetch_computed_ownership(a)
-        a.computed_regional_districts = fetch_computed_regional_districts(a)
+        fetch_computed_elevation_m(a)
```
plasticviking: Functionally looks good. I'd rather see these return a dataclass or dict or something, or atleast be named in such a way that it's obvious that they have side effects.

